PR #6, opened 2018-12-12
Fix Maven rules to support jarjar_library targets

vmax:
Adapt to using `jarjar_library`

### DIFF
--- a/maven/templates/rules.bzl
+++ b/maven/templates/rules.bzl
@@ -1,23 +1,5 @@
 # Do not edit. deployment_rules_builder.py autogenerates this file from deployment/maven/templates/rules.bzl
 
-#
-# GRAKN.AI - THE KNOWLEDGE GRAPH
-# Copyright (C) 2018 Grakn Labs Ltd
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Affero General Public License as
-# published by the Free Software Foundation, either version 3 of the
-# License, or (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


maven/templates/rules.bzl
```diff
@@ -171,7 +174,12 @@ def _deploy_maven_jar_impl(ctx):
     _generate_deployment_script(ctx, maven_coordinates)
 
     # there is also .source_jar which produces '.srcjar'
```
haikalpribadi: what's the significance of this comment?

maven/templates/rules.bzl
```diff
@@ -1,3 +1,5 @@
+# Do not edit. deployment_rules_builder.py autogenerates this file from deployment/maven/templates/rules.bzl
+
 #
 # GRAKN.AI - THE KNOWLEDGE GRAPH
 # Copyright (C) 2018 Grakn Labs Ltd
```
haikalpribadi: Linking the comment made in the other PR https://github.com/graknlabs/grakn/pull/4703#discussion_r241356298.

maven/templates/rules.bzl
vmax: @haikalpribadi this should give a hint why we use `.class_jar` in https://github.com/vmax/graknlabs-deployment/blob/master/maven/templates/rules.bzl#L174 as opposed to `.source_jar`
Any better way to phrase it?

maven/templates/rules.bzl
vmax: @haikalpribadi I'll remove license header for now

PR #7, opened 2018-12-13
Rewrite github deployment script in Python

vmax:


github/deployment.py
```diff
@@ -0,0 +1,90 @@
...
+    if len(ghr) != 1:
+        print('There should be exactly one zip archive containing `ghr`')
+        sys.exit(1)
+    subprocess.call(['unzip', ghr[0], '-d', tempdir])
+    ghr = glob.glob(os.path.join(tempdir, '**/ghr'))
+    if len(ghr) != 1:
+        print('There should be exactly one executable named `ghr` in archive')
+    ghr = ghr[0]
+elif system == 'Linux':
+    ghr = glob.glob('external/ghr_linux_tar/file/*.tar.gz')
```
lolski: @vmax would it makes sense to be explicit and just refer to the GHR tar directly here:

```
ghr = glob.glob('external/ghr_osx_zip/file/<ghr-tar-filename.tar.gz>')
```

github/deployment.py
```diff
@@ -0,0 +1,90 @@
...
+    if len(ghr) != 1:
+        print('There should be exactly one tar archive containing `ghr`')
+        sys.exit(1)
+    subprocess.call(['tar', '-xf', ghr[0], '-C', tempdir])
+    ghr = glob.glob(os.path.join(tempdir, '**/ghr'))
+    if len(ghr) != 1:
```
lolski: is this sanity check really necessary? what are the chances that we have more than one 'ghr' folder? i'm thinking if we can drop the check (trading off defensiveness for clarity)

github/deployment.py
```diff
@@ -0,0 +1,90 @@
...
+_, github_user, github_token = sys.argv
+
+with open('VERSION') as version_file:
+    distribution_version = version_file.read().strip()
+    github_tag = 'v{}'.format(distribution_version)
+
+system = platform.system()
+tempdir = tempfile.mkdtemp()
+if system == 'Darwin':
+    ghr = glob.glob('external/ghr_osx_zip/file/*.zip')
```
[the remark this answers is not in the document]
vmax: @lolski I think it's actually better to **not** specify exact filename -- it allows for easier upgrades i.e. when `ghr` releases an update you just update link in your repo but rules are still able to find the file

github/deployment.py
vmax: @lolski ^ see above

github/deployment.py
vmax: @lolski we can drop it 👌

vmax: https://github.com/graknlabs/deployment/pull/7/commits/eb4c1b2a2ca55206f6f73cb618ec85f28173e0b4

github/deployment.py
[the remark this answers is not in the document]
lolski: i see, makes sense. it is already quite safe since it's already located within its own folder `ghr_osx_zip`

github/deployment.py
lolski: i see, makes sense. it is already quite safe since it's already located within its own folder `ghr_linux_tar`

PR #10, opened 2018-12-24
Fix 'depset_is_not_iterable' incompatible change

vmax:
Fixes future-incompatible change: `depset` will be [no longer iterable](https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#depset-is-no-longer-iterable) in future version of `bazel`


vmax:
@haikalpribadi added the description

PR #11, opened 2018-12-29
Rules for Debian distribution

vmax:
This PR adds the following:
* `java_deps` rule which takes in `java_binary` and puts it together will all dependencies into a `.tgz` archive at directory `java_deps_root`. This is only meant for internal consumption of `pkg_deb`
* `deploy_deb` macro which builds final `.deb`:
```
deploy_deb(
    name = "X"
)
```

Resulting Debian package will be called `X`, target to build it is `//:X-deb`

### DIFF
--- a/distribution/java_deps.py
+++ b/distribution/java_deps.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 #
 # GRAKN.AI - THE KNOWLEDGE GRAPH
 # Copyright (C) 2018 Grakn Labs Ltd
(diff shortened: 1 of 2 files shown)


lolski:
At one point, we need to make `deploy_deb` to actually deploy to a Debian repository, which we define in `deployment.properties` like so, right @vmax ?

```
debian.repository-url=https://a.url.to.a.debian.repository.com
```

vmax:
@lolski this is right. currently `deploy_deb` only _builds_ that deb

PR #12, opened 2018-12-31
Reimplement deployment rule as a macro

vmax:
* Renames `distribution` (rule) and `_distribution_impl` (rule implementation) to `old_distribution` and old_distribution_impl`. Those are subject to removal.
* Adds `tgz2zip` rule, which converts `tar` archives to `zip`
* `java_deps` now only takes single target
* Reimplement `distribution` as a macro, which expands into sequence `java_deps` -> `pkg_tar` -> `tgz2zip`

### DIFF
--- a/distribution/BUILD
+++ b/distribution/BUILD
@@ -16,4 +16,4 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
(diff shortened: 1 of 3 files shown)


vmax:
Don't review, will be replaced by a new PR

PR #13, opened 2019-01-02
Impl. Linux packaging rules, reimpl. distribution rule

vmax:
* Renames `distribution` (rule) and `_distribution_impl` (rule implementation) to `old_distribution` and `old_distribution_impl`. Those are subject to removal.
* Adds `tgz2zip` rule, which converts `tar` archives to `zip` (used in new `distribution` macro)
* `java_deps` now only takes single target and does _not_ generate `.py` file to execute which caused collisions
* Reimplement `distribution` as a macro, which expands into sequence `java_deps` -> `pkg_tar` -> `tgz2zip`
* Rename `deploy_deb` to `linux_distributions` — macro, which expands into sequence `java_deps` -> `pkg_tar` -> {`pkg_deb`, `pkg_rpm`}

### DIFF
--- a/distribution/rules.bzl
+++ b/distribution/rules.bzl
@@ -89,23 +89,22 @@ def _tgz2zip_impl(ctx):
 
     return DefaultInfo(data_runfiles = ctx.runfiles(files=[ctx.outputs.zip]))
 
-def linux_packages(package_name,
-                   installation_dir,
-                   maintainer,
-                   version_file,
-                   description,
-                   rpm_spec_file,
-                   postinst = None,
-                   prerm = None,
-                   target = None,
-                   empty_dirs = [],
-                   files = {},
-                   depends = [],
-                   symlinks = {},
-                   modes = {}):
+
+def deb_package(name,
+                package_name,
+                installation_dir,
+                maintainer,
+                version_file,
+                description,
+                target = None,
+                empty_dirs = [],
+                files = {},
+                depends = [],
+                symlinks = {},
+                permissions = {}):
     java_deps_tar = []
     if target:
-        java_deps_name = "_{}-deps".format(package_name)
+        java_deps_name = "_{}-deb-deps".format(package_name)
         java_deps(
             name = java_deps_name,
             target = target,
@@ -113,7 +112,7 @@ def linux_packages(package_name,
         )
         java_deps_tar.append(java_deps_name)
 
-    tar_name = "_{}-tar".format(package_name)
+    tar_name = "_{}-deb-tar".format(package_name)
 
     pkg_tar(
         name = tar_name,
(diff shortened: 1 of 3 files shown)


distribution/rules.bzl
```diff
@@ -16,38 +16,30 @@
...
-        executable = java_deps_builder
+        arguments = [str(names), ctx.outputs.distribution.path],
+        executable = ctx.file._java_deps_builder,
+        progress_message = "Generating tarball with Java deps: {}".format(
+            ctx.outputs.distribution.short_path)
     )
 
 
-def _distribution_impl(ctx):
+def _old_distribution_impl(ctx):
```
lolski: is this still used? if it's just a dead-code, let's remove it.

distribution/rules.bzl
```diff
@@ -86,57 +78,152 @@ def _distribution_impl(ctx):
...
+                permissions = {}):
     java_deps_tar = []
     if target:
+        java_deps_name = "_{}-deb-deps".format(package_name)
         java_deps(
-            name = "_{}-deps".format(name),
-            targets = [target],
+            name = java_deps_name,
+            target = target,
             java_deps_root = "services/lib/"
```
lolski: what is `java_deps_root` and why is it hard-coded to `services/lib/`

distribution/rules.bzl
```diff
@@ -86,57 +78,152 @@ def _distribution_impl(ctx):
...
+        java_deps_tar.append(java_deps_name)
+
+    tar_name = "_{}-rpm-tar".format(package_name)
+    rpm_version_file = "_{}-rpm-version".format(package_name)
+
+    native.genrule(
+        name = rpm_version_file,
+        srcs = [version_file],
+        outs = [rpm_version_file.replace("-version", ".VERSION")],
+        cmd = "sed -e 's|-|_|g' $< > $@"
```
lolski: let's add an explanation with an inline comment next to the `sed` command:
```
# replaces any `-` with `_` since RPM does not allow a version number containing a `-` such as `1.5.0-SNAPSHOT`

File: distribution/rules.bzl [hunk 4]
```diff
@@ -175,7 +283,7 @@ distribution = rule(
             default="//distribution:archiver.py"
         )
     },
-    implementation = _distribution_impl,
+    implementation = _old_distribution_impl,
```
lolski: apparently `_old_distribution_impl` is used here. why is it not just called `distribution_impl`?

distribution/rules.bzl
vmax: @lolski ok

distribution/rules.bzl
vmax: @lolski
>what is java_deps_root

folder inside the `tar` archive to put JARs into

>why is it hard-coded to services/lib/

that's the sensible default for `grakn-core-server`/`grakn-core-console`

lolski: makes sense

PR #15, opened 2019-01-04
Fix rpm rules on mac

vmax:
* Import [`rpm/make_rpm.py`](https://github.com/bazelbuild/bazel/blob/760b123cdbb450c9463744724383c88e535f6ddc/tools/build_defs/pkg/make_rpm.py), [`rpm/rpm.bzl`](https://github.com/bazelbuild/bazel/blob/f09429c201347f306954479d6ad156efa7cc49d8/tools/build_defs/pkg/rpm.bzl), [`rpm/BUILD`](https://github.com/bazelbuild/bazel/blob/df16c77a501ba634b30a1e3d6f76023ac9ea93a9/tools/build_defs/pkg/BUILD) from [bazelbuild/bazel](https://github.com/bazelbuild/bazel)
* Add `rpmbuild` which specifies path to system-wide available `rpmbuild`
* Automatically select `rpmbuild` path based on platform in `pkg_rpm` macro


distribution/rules.bzl
```diff
@@ -142,12 +142,34 @@ def deploy_rpm(name,
...
+        )
+
     pkg_rpm(
         name = name,
         architecture = "x86_64",
         spec_file = spec_file,
         version_file = rpm_version_file,
         data = [tar_name],
+        rpmbuild_path = select({
+            ":linux_build": "/usr/bin/rpmbuild",
```
lolski: can we make it so it uses whatever `rpmbuild` binary which exists in our `$PATH` instead of specifying an absolute value like this? eg., if the user can invoke `rpmbuild` from his/her command line, Bazel will have access to it.

lolski:
Are we copy/pasting the `rpm` module from Bazel?

vmax:
@lolski we are

distribution/rules.bzl
vmax: @lolski we **cannot**, `bazel` forces `PATH` to be exactly `/bin:/usr/bin`

lolski:
LGTM, just two things from the discussion:
1. Separate the folder of files which we vendor-in from Bazel: `pkg_rpm_modified_from_bazel`
2. Keep the Readme files which describes the intention that we want

PR #23, opened 2019-01-09
Add no-remote tag for pkg_rpm

vmax:
`pkg_rpm` depends on having `rpmbuild` on system it is being built on, which prevents it from executing successfully on RBE


vmax:
RBE does not seem to respect `no-remote` tag at all ☹️

Issue #24, opened 2019-01-10
Create `deploy_deb` and `deploy_rpm` which deploy already built DEB and RPM package s


lolski:
A user should be able to deploy DEB and RPM using new rule `deploy_deb` and `deploy_rpm`. The rules take in already built DEB and RPM packages which you can generate using the `distribution_deb` and `distribution_rpm` rules.

To illustrate how people should use `deploy_deb` and `deploy_rpm`:
```
distribution_deb(
  name = "distribution-deb",
  ...
)
deploy_deb(
  name = "deploy-deb",
  distribution = "distribution-deb"
)

distribution_rpm(
  name = "distribution-deb",
  ...
)
deploy_rpm(
  name = "deploy-deb",
  distribution = "distribution-deb"
)
```

This is how a user can perform the deployment:
```
bazel run //:deploy-deb 
bazel run //:deploy-rpm
```

You should be able to specify to what server you want to deploy to by configuring the URL in `deployment.properties`:
```
deb.repository-url.pypi=...
deb.repository-url.test=...
rpm.repository-url.pypi=...
rpm.repository-url.test=...

```

PR #28, opened 2019-01-15
Rules for deploying RPM packages (#24)

vmax:


rpm/templates/deploy.sh
```diff
@@ -0,0 +1,51 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+#!/usr/bin/env bash
+
+set -ex
+pwd
```
lolski: do we need `pwd`?

PR #74, opened 2019-03-13
Accept credentials from environment variable in deployment rules

vmax:

`deploy_(apt|rpm|npm|pip)` rules now accept credentials via environment variables: `$DEPLOYMENT_REPO_TYPE`, `$DEPLOYMENT_USERNAME`, `$DEPLOYMENT_PASSWORD`, `$DEPLOYMENT_EMAIL` [for `npm`]

This is done so we're able to not expose our credentials (`bazel` prints out command being executed and the behaviour is non-configurable)

- Remove `$#` check for number of arguments
- If command line argument was not passed, default to environment variable (see above); if it's also not present, default to `notset`, see [this guide](https://unix.stackexchange.com/a/62337)
- Check that credential is not set to "notset`; exit with error if it is

### DIFF
--- a/apt/templates/deploy.sh
+++ b/apt/templates/deploy.sh
@@ -20,17 +20,23 @@
 
 set -e
 
-if [[ $# -ne 3 ]]; then
-    echo "Should pass <test> <deb-username> <deb-password> as arguments"
-    exit 1
-fi
 
-DEB_REPO_TYPE="$1"
-DEB_USERNAME="$2"
-DEB_PASSWORD="$3"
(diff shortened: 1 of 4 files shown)


lolski:
Is this how you supply the environment variables to the deployment command?
```
DEPLOYMENT_USERNAME=... DEPLOYMENT_PASSWORD=... DEPLOYMENT_REPO_TYPE =... bazel run //:deploy-apt
```

vmax:
@lolski right, this should be it

PR #83, opened 2019-03-25
Allow deploy_github rule to deploy releases with no files

vmax:

[graknlabs/client-python](https://github.com/graknlabs/client-python), [graknlabs/client-nodejs](https://github.com/graknlabs/client-nodejs), [graknlabs/graql](https://github.com/graknlabs/graql) all require GitHub releases with no files (actual artifacts are deployed to PyPI, NPM and Maven, respectively)

- Add a filtering statement so that we have `[]` (empty list) instead of `['']` (a list with an empty string)
- Create a dummy directory inside directory without artifacts to allow `ghr` to proceed uploading

### DIFF
--- a/github/deploy.py
+++ b/github/deploy.py
@@ -41,7 +41,7 @@ def parse_deployment_properties(fn):
     return deployment_properties
 
 
-targets = "{targets}".split(',')
+targets = list(filter(None, "{targets}".split(',')))
 has_release_description = bool(int("{has_release_description}"))
 
 properties = parse_deployment_properties('deployment.properties')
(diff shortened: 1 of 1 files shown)


github/deploy.py
```diff
@@ -41,7 +41,7 @@ def parse_deployment_properties(fn):
     return deployment_properties
 
 
-targets = "{targets}".split(',')
+targets = list(filter(None, "{targets}".split(',')))
```
lolski: I am proposing this one as I think it's less terse (especially considering that most of us here aren't Python programmers:

```
targets = [] if "{targets}".split(',') == [''] else "{targets}".split(',')
```

vmax: @lolski we can simplify it as:
`targets = [] if "{targets}" else "{targets}".split(',')`

lolski: that is even better @vmax , well done

PR #91, opened 2019-03-27
deploy_github - uploaded distribution should contain version numbers

lolski:

Add version to the filename of distributions that are uploaded to GitHub:

- grakn-core-server-windows-1.5.0.zip
- grakn-core-console-mac-1.5.0.zip
- grakn-core-server-1.5.0.tar.gz
- grakn-core-console-1.5.0.tar.gz
- grakn-core-all-mac-1.5.0.zip
- grakn-core-all-windows-1.5.0.zip
- grakn-core-console-windows-1.5.0.zip
- grakn-core-all-linux-1.5.0.tar.gz
- grakn-core-server-mac-1.5.0.zip

### DIFF
--- a/github/deploy.py
+++ b/github/deploy.py
@@ -91,8 +91,15 @@ directory_to_upload = tempfile.mkdtemp()
 dummy_directory = tempfile.mkdtemp(dir=directory_to_upload)
 
 for fl in targets:
-    filename, extension = os.path.splitext(fl)
-    final_name = '{}-{}{}'.format(filename, distribution_version, extension)
+    if fl.endswith('zip'):
+        extension = 'zip'
+    elif fl.endswith('tar.gz'):
+        extension = 'tar.gz'
+    else:
+        raise ValueError('This file is neither a zip nor a tar.gz: {}'.format(fl))
+
+    filename = fl[:-len(extension)-1]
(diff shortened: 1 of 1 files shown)


github/deploy.py
```diff
@@ -89,8 +89,18 @@ def get_github_token():
...
 dummy_directory = tempfile.mkdtemp(dir=directory_to_upload)
+
 for fl in targets:
-    shutil.copy(fl, os.path.join(directory_to_upload, os.path.basename(fl)))
+    if fl.endswith('zip'):
+        extension = 'zip'
+    elif fl.endswith('tar.gz'):
+        extension = 'tar.gz'
+    else:
+        raise ValueError('This file is neither a zip nor a tar.gz: {}'.format(fl))
```
vmax: @lolski given `bazel` limits what we pass it targets, is it even possible to hit this?

vmax: @lolski also, a better way to split out extension is `os.path.splitext`

lolski: As we've discussed in Slack,

#1 I think it's better to keep the `else` bit since the information regarding the allowed extension being `zip` and `tar.gz` is in a different file and therefore is too far removed

#2 won't work because for example, it will split `grakn-core-all.tar.gz` into `grakn-core-all.tar` + `.gz` rather than `grakn-core-all` + `.tar.gz`.

PR #115, opened 2019-04-17
Implement *_packer rules

vmax:

Have `assemble_packer` and `deploy_packer` rules useful in deployment to cloud providers, such as AWS and GCP.

The rule is essentially a Bazelised version of Hashicorp's [Packer](packer.io).

Implemented `assemble_packer` macro (based on `pkg_tar` provided by `bazel`) and `deploy_packer` rule

Sample usage:

```
assemble_packer(
    name = "assemble-gcp",
    config = ":gcp.json",
    files = {
        "//:assemble-linux-targz": "grakn-core-all-linux.tar.gz",
        …
    }
)

deploy_packer(
    name = "deploy-gcp",
    target = ":assemble-gcp",
)
```

### DIFF
--- a/packer/dependencies.bzl
+++ b/packer/dependencies.bzl
@@ -18,7 +18,7 @@
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-def deploy_gcp_dependencies():
+def deploy_packer_dependencies():
     http_archive(
          name = "packer_osx",
          url = "https://releases.hashicorp.com/packer/1.4.0/packer_1.4.0_darwin_amd64.zip",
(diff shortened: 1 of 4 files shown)


packer/templates/BUILD
```diff
@@ -0,0 +1,19 @@
+#
```
lolski: Let's remove this License as Bazel distribution shouldn't be AGPL

packer/dependencies.bzl
```diff
@@ -0,0 +1,34 @@
+#
```
lolski: Let's remove this license since Bazel-distribution shouldn't be AGPL

PR #120, opened 2019-04-19
java_deps: allow inclusion of libraries with duplicate names

### LINKED ISSUE #111: java_deps - The JARs produced by two Bazel targets from different packages but with the same label may overwrite each other
The following example demonstrates how the JARs from `//bootup:library` and `//server:library` may overwrite each other because they both produce a JAR with an identical name `liblibrary.jar`.

```
java_library(
    name = "server",
    runtime_deps = [
        "//bootup:library",
        "//server:library",
    ]
)

java_deps(
    name = "server-deps",
    target = ":server",
    java_deps_root = "server/services/lib/",
    version_file = "//:VERSION",
    visibility = ["//:__pkg__"]
)

assemble_zip(
    name = "assemble-mac-zip",
    targets = ["//:all-deps"],
)
```

The proposed solution: use the Bazel package name as a way to create namespace: `bootup-liblibrary.jar` and `server-liblibrary.jar`

vmax:

Fixes #111 

- Remove check for duplicated library file names
- Prefix every library in the archive with its owner's package name

Considering the example similar to one in the issue:

```
java_library(
    name = "xserver",
    runtime_deps = [
        "//daemon:library",
        "//server:library",
    ]
)

java_deps(
    name = "xserver-deps",
    target = ":xserver",
    java_deps_root = "server/services/lib/",
    version_file = "//:VERSION",
    visibility = ["//:__pkg__"]
)
```

these libraries with duplicate library file names will be placed in the archive:
```
-r-xr-xr-x  0 vmax   wheel    5440 Jan  1  1970 server/services/lib/daemon__liblibrary-jar.jar
-r-xr-xr-x  0 vmax   wheel   31305 Jan  1  1970 server/services/lib/server__liblibrary-jar.jar
```

### DIFF
--- a/common/rules.bzl
+++ b/common/rules.bzl
@@ -92,18 +92,14 @@ _transitive_collect_maven_coordinate = aspect(
 def _java_deps_impl(ctx):
     names = {}
     files = []
-    filenames = []
 
     mapping = ctx.attr.target[TransitiveJarToMavenCoordinatesMapping].mapping
 
     for file in ctx.attr.target.data_runfiles.files.to_list():
-        if file.basename in filenames:
-            continue # do not pack JARs with same name
         if file.extension == 'jar' and not file.path.startswith(LOCAL_JDK_PREFIX):
-            filename = mapping.get(file.path, file.basename).replace('.', '-').replace(':', '-')
+            filename = file.owner.package.replace('/', '-') + '__' + mapping.get(file.path, file.basename).replace('.', '-').replace(':', '-')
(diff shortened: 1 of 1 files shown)


haikalpribadi:
For a `java_library` target that exists in `@graknlabs_grakn_core//common:common`, the bundled jar should be named: `graknlabs-grakn-core-common-common.jar`

vmax:
alternative approach is taken here: #139

PR #122, opened 2019-04-25
assemble_maven and deploy_maven should produce and upload Maven artifact with the correct pom files

lolski:

There are two `pom.xml`s that got produced, the actual `pom.xml` file and an additional copy which is embedded inside the JAR which servers as metadata.

During the assembly step, these two files should just contain a version placeholder `{pom_version}` rather than the actual version number. The actual version number would only be added during the deployment step.

### DIFF
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -29,7 +29,6 @@ import re
 import subprocess as sp
 import sys
 import tempfile
-
 import zipfile
 
 
@@ -46,6 +45,7 @@ class ZipFile(zipfile.ZipFile):
         attr = member.external_attr >> 16
         os.chmod(ret_val, attr)
         return ret_val
+
 
 def parse_deployment_properties(fn):
     deployment_properties = {}
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -30,8 +30,22 @@ import subprocess as sp
 import sys
 import tempfile
 
-print(os.getcwd())
+import zipfile
 
+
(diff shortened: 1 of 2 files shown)


maven/templates/deploy.py
```diff
@@ -30,9 +30,23 @@
 import sys
 import tempfile
 
-print(os.getcwd())
+import zipfile
 
 
+# This ZipFile extends Python's ZipFile and fixes the lost permission issue
```
vmax: @lolski we don't really have permissions issue while repacking JARs — there are no executables with `+x` bit

maven/templates/deploy.py
```diff
@@ -54,6 +68,30 @@ def md5(fn):
     return hashlib.md5(open(fn).read()).hexdigest()
 
 
+def update_pom_within_jar(jar_path, new_pom_content):
+    ext = 'jar'
+    original_jar_basename = os.path.basename(jar_path[:-len(ext) - 1])
+    updated_jar_basename = original_jar_basename + '-updated'
```
vmax: @lolski make sure we keep the old names in final version

maven/templates/deploy.py
lolski: Good point, addressed in a911aea04576669bfeae464158eb457613f1f20a

maven/templates/deploy.py
lolski: Good point, addressed in a911aea04576669bfeae464158eb457613f1f20a

Issue #124, opened 2019-05-02
Create assemble + deploy AWS and assemble + deploy GCP rules on top of assemble + deploy Packer


lolski:
The main motivation is because `deploy_packer` does not follow the same UI / UX of other rules such as `deploy_apt` and `deploy_rpm`: we need to have two Packer configurations, `snapshot.json` and `release.json`, instead of having the configuration defined in `deployment.properties`.

From the users' point of view the new rule should be configured and executed in this way:
```
# deployment.properties
repo.gcp.snapshot=grakn-dev
repo.gcp.release=grakn-public
repo.aws.snapshot=...
repo.aws.release=...

# running
$ bazel run //:deploy-gcp -- {snapshot|release}
$ bazel run //:deploy-aws -- {snapshot|release}
```

PR #128, opened 2019-05-10
Add assemble_gcp rule that encapsulates generating config

vmax:

First part of making AWS/GCP rules more user-friendly (issue #124)

- Adds `generate_json_config` rule which allows to substitute values in template
- Adds `assemble_gcp` which generates config and calls `assemble_packer`. Sample usage:

```
assemble_gcp(
    name = "assemble-gcp-snapshot",
    files = GCP_FILES,
    install = "//deployment/deploy-gcp-image/files:install.sh",
    project_id = "grakn-dev",
    zone = "europe-west1-b",
    image_name = "grakn-kgms-snapshot-{{user `version`}}",
    image_licenses = 'projects/grakn-public/global/licenses/grakn-kgms-premium'
)
```

### DIFF
--- a/common/generate_json_config.bzl
+++ b/common/generate_json_config.bzl
@@ -1,0 +1,38 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


lolski:
@vmax right now, does the script expect a file `install.sh` to always be included in `files`?

vmax:
@lolski it does

lolski:
```
assemble_gcp(
    name = "assemble-gcp-snapshot",
    files = GCP_FILES,
    project_id = "grakn-dev",
    zone = "europe-west1-b",
    image_name = "grakn-kgms-snapshot-{{user `version`}}",
    image_licenses = 'projects/grakn-public/global/licenses/grakn-kgms-premium',
    install = "//:deployments/gcp/install.sh"
)
```

what if we make it explicit like this @vmax ^

PR #129, opened 2019-05-13
Add assemble_aws rule that encapsulates generating config

vmax:

Second part of making AWS/GCP rules more user-friendly (issue #124)

- Adds `assemble_aws` which generates config and calls `assemble_packer`. Sample usage:

```
assemble_aws(
    name = "assemble-aws-snapshot",
    ami_name = "grakn-kgms-snapshot-{{user `version`}}",
    install = "//deployment/deploy-aws-image/files:install.py",
    region = "us-east-1",
    files = AWS_FILES
)
```

### DIFF
--- a/aws/rules.bzl
+++ b/aws/rules.bzl
@@ -1,0 +1,44 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


aws/rules.bzl
```diff
@@ -0,0 +1,44 @@
...
+                 install,
+                 region,
+                 files):
+    install_fn = Label(install).name
+    generated_config_target_name = name + "__do_not_reference_config"
+    generate_json_config(
+        name = generated_config_target_name,
+        template = "@graknlabs_bazel_distribution//aws:packer.template.json",
+        substitutions = {
+            "{ami_name}": ami_name, # grakn-kgms-{{user `version`}}
```
lolski: Is this comment needed or can be removed? `# grakn-kgms-{{user `version`}}`

vmax: @lolski thanks for noticing, that should be removed

vmax: @lolski fixed in https://github.com/graknlabs/bazel-distribution/pull/129/commits/377ea4c208f0129f2aed1c95a9f46bf0190c9d78

PR #134, opened 2019-06-12
Fix Maven deployment: should be allowed without GPG key

vmax:

Recent addition of signing Maven artifacts (#133) broke existing deployments since they don't satisfy `gpg key should already be present in keyring` requirement. This PR allows to deploy artifacts without signature if signing process fails for any reason.

If signing process fails, `None` is returned instead of throwing an error, which results in _not_ uploading signature file (which is the same behaviour as previously)

### DIFF
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -90,20 +90,24 @@ def upload(url, username, password, local_fn, remote_fn):
             local_fn, upload_status_code))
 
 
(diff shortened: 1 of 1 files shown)


vmax:
After discussion with @lolski we decided to pick alternative approach: user of the rule would choose it he wants to sign his artifacts by supplying optional command line arguments.

PR #136, opened 2019-06-17
Fix an issue where assemble-maven and deploy-maven would incorrectly use project name and description as group and artifact id

lolski:

This PR fixes an issue where assemble-maven and deploy-maven would incorrectly use project name and description as group and artifact id. This is caused by the fact that the process of parsing `groupId`, `artifactId`, and `version` relied on brittle array indexing where we're expecting them to be element number 1, 2,and 3 in the `pom.xml`.

This logic has been made robust by retrieving the element by name, ie., `root.find('groupId')`.


maven/assemble.py
```diff
@@ -30,9 +30,11 @@
...
 _, output_jar_path, input_jar_path, pom_file_path = sys.argv
 
-pom_file_tree = ElementTree.parse(pom_file_path)
-group_id, artifact_id, version = list(map(attrgetter('text'),
-                                          pom_file_tree.getroot()[1:4]))
+namespace = { 'namespace': 'http://maven.apache.org/POM/4.0.0' }
+root = ElementTree.parse(pom_file_path).getroot()
+group_id = root.find('namespace:groupId', namespace).text
+artifact_id = root.find('namespace:artifactId', namespace).text
+version = root.find('namespace:version', namespace).text
```
vmax:
```suggestion
if not groud_id:
    raise Exception("could not get groupId from POM")
  …
```

PR #168, opened 2019-07-15
Pass tags to pkg_tar

prestonvanloon:

To pass the `tags` property to pkg_tar. The motivation is that I want to signal to our CI runner to skip this in normal presubmit builds.

```python
tags = ["skipci"],
```

Then this target wouldn't be built as part CI. 

This PR passes the `tags` property to pkg_tar as part of assemble_targz.


vmax:
@prestonvanloon thank you for your contribution!

PR #169, opened 2019-07-16
Add tags to genrule for assemble_targz

prestonvanloon:
Follow up to #168, I forgot to pass tags to another rule within assemble_targz


vmax:
@prestonvanloon thank you for your contribution!

PR #170, opened 2019-07-18
Prepare for upgrading of @build_bazel_rules_nodejs

vmax:

As migration to latest `@build_bazel_rules_nodejs` is planned, we need to adapt `assemble_npm`/`deploy_npm` to newest rules [while also preserving compatibility with currently-used version].

- Use `@nodejs//:npm` (compatible with both old and new `rules-nodejs`)
- Append `external/nodejs_<platform>_amd64/bin/` to `PATH` when invoking `npm`. Adding `/bin/` and `/usr/bin` is also required to have `bash`/`dirname` available (needed by `npm` script)

### DIFF
--- a/npm/assemble.py
+++ b/npm/assemble.py
@@ -66,7 +66,14 @@ subprocess.check_call([
     'npm',
     'pack'
 ], env={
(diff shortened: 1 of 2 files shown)


lolski:
LGTM

PR #186, opened 2019-09-25
Fix deploy_maven rule to be compatible with Python 3

vmax:

`deploy_maven` rule broke after we upgraded to `bazel 0.29.1` (and running it with Python 3) with the following exception:
```
  File "/home/circleci/.cache/bazel/_bazel_circleci/98db3734c28662e6d9e008d213a17afe/execroot/graknlabs_graql/bazel-out/k8-fastbuild/bin/grammar/deploy.py", line 163, in <module>
    pom_updated.write(pom_updated_content)
  File "/opt/circleci/.pyenv/versions/3.5.2/lib/python3.5/tempfile.py", line 483, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'
```

Open all temporary files in write-text mode (`wt`) so we could write `str` to it (without first encoding to `bytes`)

### DIFF
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -158,7 +158,7 @@ filename_base = '{coordinates}/{artifact}/{version}/{artifact}-{version}'.format
 pom_updated = None
 jar_updated = None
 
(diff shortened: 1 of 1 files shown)


lolski:
LGTM

Issue #150, opened 2019-07-03
Streamline how VERSION is provided for all assembly and deployment rules


haikalpribadi:
At the moment, we have different ways to VERSION our distribution for assembly and deployment. 

For example, for `assemble_rpm` and `deploy_rpm`, we strictly rely on the content of the `//:VERSION` file, so we overwrite the file before calling assembling or deploying.
```
echo $(cat VERSION)_$CIRCLE_SHA1 > VERSION && cat VERSION
bazel run //:deploy-rpm -- snapshot
```

For `assemble_maven` and `deploy_maven`, we repackage the JAR at deployment with the version we want:
```
bazel run //common:deploy-maven -- snapshot $CIRCLE_SHA1
bazel run //common:deploy-maven -- release $(cat VERSION)
```

This is very inconsistent, and in cases where we need to overwrite the `//:VERSION` file, it's actually problematic/error-prone: what if we a developer forgets to update the `//:VERSION` file and calls `bazel run //:deploy-rpm -- snapshot`, because it's not one canonical step?

Let's streamline the way version our distribution for all assembly and deployment rules as described below.

1) Assembly rules do not take in `version_file = "//:VERSION"` any more, and can optionally be versioned using Bazel environment variables. For example:
   * `bazel build //common:assemble-maven` will produce a Maven JAR with version `0.0.0` (aka. _"unversioned"_).
   * `bazel build --define version=$GIT_SHA //common:assemble-maven` will produce Maven JAR with a git commit as the version.
   * `bazel build --define version=$(cat VERSION) //common:assemble-maven` will produce a Maven JAR with the content of `//:VERSION` (e.g. `1.5.7`) as the version.

2) Deployment rules remain to be parameterised with `release` or `snapshot`, but do not take in any version arguments any more. They will verify the distribution they receive from assembly rules is versioned properly, unique to each platform's versioning syntax requirements. For example:
   * `bazel run //common:deploy-maven -- snapshot` will make sure that the Maven jar is not versioned `0.0.0` (i.e. not _"unversioned"_)
   * `bazel run //common:deploy-maven -- snapshot` will make sure that they Maven jar is semantically version (`X.Y.Z-*`) (this just an example)
   * `bazel run //:deploy-rpm -- release` will make sure that the RPM distro is versioned `X.Y.Z-{architecture}` (this is just an example)

The above design should be applied to all `assemble_*` and `deploy_*` rules.

haikalpribadi:
[...]

PR #196, opened 2019-10-15
Refactor how version is provided to assemble_versioned

vmax:

Partial work on #150

`assemble_versioned` no longer requires user to provide `version_file`. Instead it can be specified as `bazel build --define version=$(cat VERSION) <assemble_versioned_target>`

### DIFF
--- a/common/assemble_versioned.bzl
+++ b/common/assemble_versioned.bzl
@@ -18,13 +18,25 @@
 #
 
 def _assemble_versioned_impl(ctx):
+    if not ctx.attr.version_file:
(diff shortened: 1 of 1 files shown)


lolski:
This PR adds support for `--define version=...` but keeps the ability of supplying the Version file using the `version_file` field in the rule.

I prefer to delete the latter to keep the code simple. However @vmax prefer to keep it since he argues the UX is better for people who has no need to declare the version during the invocation of `bazel build ...`.

Maybe we can get a third opinion from @haikalpribadi.

PR #206, opened 2020-01-04
Pass tags attribute to all rpm internal rules.

### LINKED ISSUE #205: assemble_rpm does not allows custom tags to be specified
`assemble_rpm` rule should allow to specify common `tags` attribute and pass it to all the wrapped rules.

This can be beneficial in the cases when tags are intensively used. This would also allow to specify `tags = ["manual"]` so that `.rpm` is ignored when running `bazel build //...` and similar for the setups where `rpmbuild` is not available.

zoido:
Adding option to pass `tags` to the `assemble_rpm` rule.
This can be beneficial in the cases when tags are intensively used. This would also allow to specify `tags = ["manual"]` so that `.rpm` is ignored when running `bazel build //...` and similar for the setups where `rpmbuild` is not available.
ref: #205

 - Adds a new `tags` argument to the `assemble_rpm` macro and pass it to all wrapped rules.

Additionally:
 - remove references to the `.deb` packages
 - add `workspace_refs` argument doc

RESOLVES #205

### DIFF
--- a/rpm/rules.bzl
+++ b/rpm/rules.bzl
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-load("@rules_pkg//:pkg.bzl", "pkg_tar", "pkg_deb")
+load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@rules_pkg//:rpm.bzl", "pkg_rpm")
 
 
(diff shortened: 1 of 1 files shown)


vmax:
@zoido thank you for your contribution!

Issue #215, opened 2020-02-16
Unique POM File Name


amit-traiana:
I own a project where a single BUILD file will define two `java_library` targets. At the moment, `build-tools` always use hard-coded POM file named `pom.xml`. So if I'm running both targets, the first `pom.xml` will get overwrited by the second. I created a fork that names the pom file based on the target name (`my_target_pom.xml` for example). 

This solve the issue but create another one - `deploy_maven` is correctly coupled into `assemble-maven` - it will look for that statically named pom file named `pom.xml`. `deploy_maven` support passing the `assemble-maven` target as optional argument which solve this issue, but it means this argument must be a must and not optional (which makes sense to me - as there is already sort of a coupling between the two with the static pom file name).

Let me know if that works for you, and if so - I'll send a pull-request.

Thanks!

amitz:
#216

vmax:
Fixed by #217

Issue #214, opened 2020-02-16
Supporting Targets with No .Srcjar


amit-traiana:
I'm using `assemble-maven` in a Scala project. [scala_library ](https://github.com/bazelbuild/rules_scala/blob/master/docs/scala_library.md) is implemented differently to `java_library` so the above if condition will not trigger:

https://github.com/graknlabs/build-tools/blob/master/distribution/maven/rules.bzl#L270-L277

So the first thing will be support `scala-rules` if possible, because both are very similar (use POM, Jar file, srcjar etc) and Scala users can also benefits from `assemble-maven`.

Second issue, that if the `target` is not of type `java`, it will revert into the second option in the if statement. The 2nd if statement never defines `srcjar` like the first option, so the Depset will fail to populate because the variable was never created. 

Thanks!

vmax:
Fixed by #217

PR #216, opened 2020-02-18
Adding Scala and multiple POM files support.

amitz:

Fixing three  issues:

1. I'm also using Scala here, which is almost identical in deployment to Java. However, using [scala_library](https://github.com/bazelbuild/rules_scala/blob/master/docs/scala_library.md) with `assemble_maven` doesn't work. 

2. The POM file name is hard-coded (`pom.xml`). Because we use both Scala and Java in the same BUILD file, the file was getting overwrite. #215 

3. Targets that have only the `files` attribute failed to run because the provider return `srcjar` when there's isn't any defined. #214 

- `pom.xml` will now be named based on the target name. So if my `assemble_maven` target name is `hello_world`, the pom is called `hello_world_pom.xml`.

- Side effect of the above is that now `deploy_maven` requires a target (it was optional). The reason for that is because `pom.xml` was hard-coded into `deploy_maven`. Now the unique pom file name is being grabbed from the Target directly.

- I changed the if condition that extracts jars and srcjars. First I'm checking the target has `files` attribute in it. If it does, I'm checking that there's files in `files`, and lastly I check the first file has JAR file in it (like you previously did). I also modified the way srcjars are being picked. I'm looking for `-src.jar` ending exclusively (the previous code was picking the first file in the list, and it wasn't the right srcjar using `scala_library` targets). 

- Srcjar is not available on all jars. In case there is no srcjar, the Provider no longer return it (it used to fail here).

## Notes
I tested it on my Bazel project. I didn't find any tests to run the code changes on, so it's probably a good idea to try to run it on your Bazel projects to make sure it didn't break anything I wasn't aware of.

Thanks.

### DIFF
--- a/maven/templates/rules.bzl
+++ b/maven/templates/rules.bzl
@@ -162,9 +162,9 @@ def _parse_maven_coordinates(coordinate_string):
 
 def _generate_pom_xml(ctx, maven_coordinates):
     # Final 'pom.xml' is generated in 2 steps
-    preprocessed_template = ctx.actions.declare_file("_pom.xml")
+    preprocessed_template = ctx.actions.declare_file("_{}_pom.xml".format(ctx.attr.name))
 
-    pom_file = ctx.actions.declare_file("pom.xml")
+    pom_file = ctx.actions.declare_file("{}_pom.xml".format(ctx.attr.name))
 
     maven_pom_deps = ctx.attr.target[MavenPomInfo].maven_pom_deps
     deps_coordinates = depset(maven_pom_deps).to_list()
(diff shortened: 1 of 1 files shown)


amit-traiana:
There's another thing to consider here it seems. When using `exports`, I had problem with:

```
target_string = target[JavaLibInfo].target_coordinates.to_list()[0]
```

The first element in the array was the dependency of the target. It seems like `[-1]` was a better option in all the scenarios I checked, but it's probably better to write a logic to handle it instead of guessing.

amitz:
There's another thing to consider here it seems. When using `exports`, I had problem with:

```
target_string = target[JavaLibInfo].target_coordinates.to_list()[-1]
```

The first element in the array was the export, and not the root target itself. It seems like `[-1]` worked better on couple of scenario I tested it again, but it's probably better to write a logic to pick the right target instead of guessing.

PR #234, opened 2020-06-18
Add deploy_distribution rule

adammitchelldev:

Since many of our tests rely on separately executable artifacts (mainly grakn) that should not share any dependencies with the tested artifact, we want to be able to deploy versioned snapshot and release artifacts built for specific commits with the correct versions of dependencies so that tests can properly run against a static version that will not change due to overlaid bazel dependencies.

This requires writing a bazel rule that will publish some archive/file to a raw repository, which is the goal of this PR.

- Added `deploy_distribution` rule that publishes a single artifact to a raw repo.

### DIFF
--- a/distribution/rules.bzl
+++ b/distribution/rules.bzl
@@ -1,3 +1,22 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
 def _deploy_distribution_impl(ctx):
     _deploy_script = ctx.actions.declare_file("{}_deploy.py".format(ctx.attr.name))
 
--- a/distribution/rules.bzl
+++ b/distribution/rules.bzl
@@ -1,0 +1,80 @@
+def _deploy_distribution_impl(ctx):
+    _deploy_script = ctx.actions.declare_file("{}_deploy.py".format(ctx.attr.name))
+
+    version_file = ctx.actions.declare_file(ctx.attr.name + "__do_not_reference.version")
+    version = ctx.var.get('version', '0.0.0')
+
+    ctx.actions.run_shell(
+        inputs = [],
+        outputs = [version_file],
+        command = "echo {} > {}".format(version, version_file.path),
+    )
+    
+    ctx.actions.expand_template(
+        template = ctx.file._deploy_script,
+        output = _deploy_script,
+        substitutions = {
+            "{version_file}": version_file.short_path,
+            "{artifact_path}": ctx.file.target.short_path,
+            "{artifact_group}": ctx.attr.artifact_group,
+        },
+    )
+    files = [
+        ctx.file.target,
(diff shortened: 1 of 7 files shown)


distribution/templates/deploy.py
```diff
@@ -0,0 +1,100 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+from __future__ import print_function
+from xml.etree import ElementTree
+
+import hashlib
```
adammitchelldev, line 25: We can definitely remove some of these imports

distribution/rules.bzl
```diff
@@ -0,0 +1,80 @@
+def _deploy_distribution_impl(ctx):
```
vmax, line 1: @adammitchelldev can we have an Apache license header here as well?

distribution/templates/deploy.py
vmax: @adammitchelldev right! I don't see `ElementTree` annd `hashlib` used anywhere. Same goes for `zipfile` and `tempfile`.

distribution/templates/deploy.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def unpack_args(_, a):
+    return a
+
+
+if len(sys.argv) < 2:
+    raise ValueError('Should pass <snapshot|release> as arguments')
+
+repo_type = unpack_args(*sys.argv)
```
vmax, line 62: @adammitchelldev consider using this:

```suggestion
_, repo_type = sys.argv
```

This would also require us to stricten the condition above to be `len(sys.argv)` to be exactly `2`

distribution/templates/deploy.py
```diff
@@ -0,0 +1,100 @@
...
+username, password = os.getenv('DEPLOY_DISTRIBUTION_USERNAME'), os.getenv('DEPLOY_DISTRIBUTION_PASSWORD')
+
+if not username:
+    raise ValueError('Error: username should be passed via $DEPLOY_DISTRIBUTION_USERNAME env variable')
+
+if not password:
+    raise ValueError('Error: password should be passed via $DEPLOY_DISTRIBUTION_PASSWORD env variable')
+
+deployment_properties = parse_deployment_properties('deployment.properties')
+distribution_url = deployment_properties['repo.distribution.' + repo_type]
```
vmax, line 73: @adammitchelldev we're accessing `repo_type` before validating its correctness

distribution/templates/deploy.py
```diff
@@ -0,0 +1,100 @@
...
+if repo_type == 'snapshot' and len(re.findall(version_snapshot_regex, version)) == 0:
+    raise ValueError('Invalid version: {}. An artifact uploaded to a {} repository '
+                     'must have a version which complies to this regex: {}'
+                     .format(version, repo_type, version_snapshot_regex))
+if repo_type == 'release' and len(re.findall(version_release_regex, version)) == 0:
+    raise ValueError('Invalid version: {}. An artifact uploaded to a {} repository '
+                     'must have a version which complies to this regex: {}'
+                     .format(version, repo_type, version_snapshot_regex))
+
+split_artifact = os.path.basename('{artifact_path}').split('.', 1)
```
vmax, line 95: @adammitchelldev 
```suggestion
split_artifact = os.path.splitext('{artifact_path}')
```

distribution/rules.bzl
adammitchelldev: Will do

distribution/templates/deploy.py
adammitchelldev: @vmax Actually this doesn't quite do what I'd want, which is to work with `.tar.gz` files. Unfortunately `splitext` splits purely from the last `.`, which would lead to me inserting the version as `.tar-{version}.gz`.

vmax: @adammitchelldev valid point! disregard this comment then

distribution/templates/deploy.py
adammitchelldev: Good spot btw

PR #240, opened 2020-07-27
maven(github packages) - return http status code 200

socar-dorma:

Fixed maven deploy to be uploadable to github package.

Also allowed when uploading http status code 200.

### DIFF
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -55,7 +55,7 @@ def upload(url, username, password, local_fn, remote_fn):
         urljoin(url, remote_fn)
     ]).decode().strip()
 
-    if upload_status_code != '200' or upload_status_code != '201':
+    if not (upload_status_code == '200' or upload_status_code == '201'):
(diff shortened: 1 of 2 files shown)


maven/templates/deploy.py
```diff
@@ -55,7 +55,7 @@ def upload(url, username, password, local_fn, remote_fn):
         urljoin(url, remote_fn)
     ]).decode().strip()
 
-    if upload_status_code != '201':
+    if not (upload_status_code == '200' or upload_status_code == '201'):
```
vmax, line 58: @socar-dorma 

```suggestion
    if upload_status_code not in {'200', '201'}:
```

vmax:
@socar-dorma I think you introduced more changes since my original review. Could you only leave those that are related to `200`/`201` http code and submit a separate PR with description on why other changes are needed?

socar-dorma:
@vmax 
sorry.
After I received the fork, I worked on the master branch and the changes were mixed.
I'll post it again after cleaning up.

PR #250, opened 2020-08-19
Fix old skydoc dependency issue

adammitchelldev:

We depend on a custom fork of skydoc, which in turn depends on a custom fork of bazel, in order to get skydoc to be able to correctly read "non-public" source dependencies of `.bzl` files from other repos. This causes an issue because skydoc (now stardoc) depends on old versions of `nodejs_rules` that are not forwards compatible, which causes our `npm` rules and anyone who depends on them to only be available for old `nodejs_rules` users.

- Updated dependency on `nodejs_rules`
- Updated dependency on `stardoc`, formerly our fork of `skydoc`
- Fixed the visibility issue by depending on sources directly in a `bzl_library` target for `stardoc` to be able to see them on some deps root.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -18,6 +18,18 @@
 #
 
 load("@io_bazel_skydoc//stardoc:stardoc.bzl", "stardoc")
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
+bzl_library(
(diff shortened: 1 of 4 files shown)


adammitchelldev:
THIS IS NOT COMPLETE

Issue #251, opened 2020-08-19
Issue with `skydoc`/`stardoc` dependency and `nodejs_rules` versions


adammitchelldev:
We forked `skydoc` (now called `stardoc`) a while ago due to an issue where Bazel can't see various dependencies that are `load(..)`ed whilst interpreting the `.bzl` files ([originally raised here](https://github.com/bazelbuild/skydoc/issues/166)). This is because it effectively hacks the way that the files are read to read from predefined roots that it kind of "guesses" will work ([it's an acknowledged issue in the `stardoc` source](https://github.com/bazelbuild/stardoc/blob/87dc99cfe1baa9255c607ac0229bfd33a65367f5/stardoc/stardoc.bzl#L42-L55)), and this breaks when loading files outside of the current WORKSPACE.

However, this causes us to depend on a very old version of `skydoc`, which in turn depends on an old version of `skylib`, which in turn depends on an old version of `rules_sass`, which depends on an old version of `nodejs_rules`, which means that our `npm_deploy` rule can only work the old `nodejs_rules` (since updating the `nodejs_rules` breaks `skydoc`). This limits our ability to depend on the latest `nodejs_rules` downstream, such as in `client-java`, which means we can't upgrade to the latest version of npm `@bazel/jasmine`. The `nodejs_rules` changes have been very major and no longer backwards compatible and it should be considered a significant tech debt if we can't upgrade.

We started trying to solve the issue and found a workaround as listed in the original issue: include the dependent source files as deps to `stardoc` via another `bzl_library` rule, however, this only works when the files themselves are public (bazel visibility wise). We fixed this with workarounds for everything except `pip_import`, which generates a bazel workspace with a `requirements.bzl` that it does not export as public. Unfortunately, this means our workaround does not fully fix everything and we cannot build docs for the `pip` rules.

This leaves us with a few solutions:

### Do nothing now and hope the issue can be fixed upstream

We don't immediately need to update `rules_nodejs` but this isn't actually a solution. There is no "clear party to blame" here that we can submit an easy fix to, the "real fix" is that the whole thing should work differently, but it may be possible to convince one of the involved repos to allow configuration for a workaround.

### Use a patch to complete the workaround

[...]

PR #252, opened 2020-08-20
Fix old dependency on Skydoc, update rules_nodejs

adammitchelldev:

We found that skydoc was forcing us to depend on an old version of rules_nodejs, preventing us from upgrading elsewhere. This issue is described in detail in #251.

The goal of this PR is to remove our dependency on forks of `skydoc` and `bazel` by applying a patch to `rules_python` that allows the most recent version of `stardoc` to work.

- Apply a patch to `rules_python` (which is now directly imported and has been updated.)
- Updated `rules_nodejs`.
- Apply the "official" workaround for `stardoc`, which is to include the sources of files that our rules depend on in the stardoc rule as source dependencies.

### DIFF
--- a/npm/rules.bzl
+++ b/npm/rules.bzl
@@ -74,7 +74,7 @@ assemble_npm = rule(
             cfg = "host"
         ),
         "_npm": attr.label(
-            default = Label("@nodejs//:node"),
+            default = Label("@nodejs//:npm"),
             allow_files = True
         )
     },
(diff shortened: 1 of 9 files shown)


vmax (approved):

npm/rules.bzl
```diff
@@ -74,7 +74,7 @@ assemble_npm = rule(
             cfg = "host"
         ),
         "_npm": attr.label(
-            default = Label("@nodejs//:bin/npm"),
+            default = Label("@nodejs//:node"),
```
adammitchelldev, line 77: This is a migration change necessary for `rules_nodejs`. Likely, the `assemble_npm` and `deploy_npm` rules need to change entirely, but we are not in a position to test this yet.

lolski (approved):

PR #253, opened 2020-08-24
Extends artifact_extractor to support zip decompression

sorsaffari:
As we may need to extract `.zip` artifacts (for Windows), the `artifact_extractor` has been extended to choose a decompression tool according to the file's extension.

- added unzip script
- the extract file extension and use  the right script to decompress it


artifact/rules.bzl
```diff
@@ -148,22 +148,43 @@ def artifact_file(name,
...
 """
 
 def _artifact_extractor_impl(ctx):
     artifact_file = ctx.file.artifact
+    artifact_extention = artifact_file.extension
+
+    if artifact_extention == 'zip': 
+        target_script_template = script_template_zip
+        artifact_unpacked_name = artifact_file.basename.replace('.zip', '')
+    elif artifact_extention == 'gz':
```
adammitchelldev, line 175: I know it's annoying but we should probably do a full check for `.tar.gz` and alternative extensions that `tar` will support, such as `.tgz`, just to make sure the rule is going to be a bit more future proof! You may have to use some starlark string matching instead of the convenient `.extension` sadly.

artifact/rules.bzl
```diff
@@ -148,22 +148,43 @@ def artifact_file(name,
...
+        target_script_template = script_template_gz
+        artifact_unpacked_name = artifact_file.basename.replace('.tar.gz', '')
+    else:
+        fail("Extention [{extention}] is not supported by artifiact_etractor".format(extention = artifact_extention))
```
adammitchelldev, line 179: Nice!

artifact/rules.bzl
```diff
@@ -148,22 +148,43 @@ def artifact_file(name,
         **kwargs
     )
 
-script_template = """\
+script_template_gz = """\
```
adammitchelldev, line 151: Let's make this `script_template_targz` to be consistent with naming in our other rules (such as `assemble_targz`)

artifact/rules.bzl
```diff
@@ -177,11 +198,7 @@ artifact_extractor = rule(
...
             doc = "Artifact archive to extract.",
-        ),
-        "strip_components": attr.int(
-            default = 2,
-            doc = "tar --strip-components argument (default 2).",
         )
     },
     executable = True,
-)
+)
```
adammitchelldev, line 204: Missing newline at end of file!

adammitchelldev (changes requested):

artifact/rules.bzl
sorsaffari: resolved in https://github.com/graknlabs/bazel-distribution/pull/253/commits/07cd24b22dad38db1a949adb58388f514abaa9e0

artifact/rules.bzl
sorsaffari: in https://github.com/graknlabs/bazel-distribution/pull/253/commits/07cd24b22dad38db1a949adb58388f514abaa9e0 I've renamed this to `script_template_tar` after its extraction method

artifact/rules.bzl
sorsaffari: https://github.com/graknlabs/bazel-distribution/pull/253/commits/844aa511bfaf967c1cdf60880c3d2e8509321fb8

adammitchelldev (approved):

PR #259, opened 2020-08-26
Makes artifact-extractor agnostic to the OS it runs on

sorsaffari:
As the `grakn-extractor` may run on different OSes, we have tweaked the implementation to support this requirement.

- implemented `artifact_extractor` as a macro that calls the previously implemented `_artifact_extractor` followed by `sh_binary`


adammitchelldev (approved):

vmax:
Looks like it's no longer needed.

PR #260, opened 2020-09-18
Make the pom_replace_version.py script a py_binary

nlewo:

Make the rule more hermetic since it currently depends of the presence of
Python in the PATH. With this patch, it would use the interpreter from the Python toolchain.

Wrap the `pom_replace_version` script in a `py_binary`. Note the `assemble` script is already built as a `py_binary`.


vmax (approved):

vmax:
@nlewo thank you for your contribution!

PR #263, opened 2020-10-02
Allow absent srcjar in deploy_maven

vmax:

Follow-up to #262 which tweaks `deploy_maven` to access artifacts without source jars

Gracefully handle absence of source JAR while deploying


alexjpwalker (approved):

lolski:
Looks good to me. Not having a source JAR is a valid case. The PR makes sense.

PR #264, opened 2020-10-06
Allow java_deps to target Maven packages and support overrides for the output path

alexjpwalker:

To allow Grakn 2.0 to output rocksdb-dev to lib/dev and rocksdb-prod to lib/prod when building a Grakn distribution.

Firstly, we need to be able to target Maven packages using `java_deps`. We add support for this by looking through both `ctx.attr.target.data_runfiles.files` and `ctx.attr.target.files`.

We also add support for `java_deps` to have output path overrides. When a JAR matching the given pattern is found, it is deployed to the folder specified in the override, instead of the default output folder.

### DIFF
--- a/common/java_deps.bzl
+++ b/common/java_deps.bzl
@@ -125,7 +125,7 @@ def _java_deps_impl(ctx):
             for jarPattern in outputPathOverrides:
                 if file.basename == jarPattern or (jarPattern.endswith("*") and file.basename.startswith(jarPattern.rstrip("*"))):
                     names[file.path] = outputPathOverrides.get(jarPattern) + filename + ".jar"
-                    continue
+                    break
             if names.get(file.path) == None:
                 names[file.path] = ctx.attr.java_deps_root + filename + ".jar"
             files.append(file)
--- a/common/java_deps.bzl
+++ b/common/java_deps.bzl
@@ -110,6 +110,7 @@ def _java_deps_impl(ctx):
     names = {}
     files = []
     filenames = []
+    outputPathOverrides = ctx.attr.java_deps_root_overrides
 
     mapping = ctx.attr.target[TransitiveJarToMavenCoordinatesMapping].mapping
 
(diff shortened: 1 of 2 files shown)


alexjpwalker:
Do not merge until an issue is fixed that causes the documentation (README.md) to not include the new field in java_deps.

common/java_deps.bzl
```diff
@@ -110,18 +110,24 @@ def _java_deps_impl(ctx):
...
             filename = mapping.get(file.path, file.basename).replace('.', '-').replace(':', '-')
             if filename in filenames:
                 print("Excluded duplicate: {}".format(filename))
                 continue # do not pack JARs with same name
-            names[file.path] = ctx.attr.java_deps_root + filename + ".jar"
+            for jarPattern in outputPathOverrides:
+                if file.basename == jarPattern or (jarPattern.endswith("*") and file.basename.startswith(jarPattern.rstrip("*"))):
+                    names[file.path] = outputPathOverrides.get(jarPattern) + filename + ".jar"
+                    continue
+            if names.get(file.path) == None:
```
grabl, line 129: @alexjpwalker 

```suggestion
            if file.path not in names:
```

vmax, line 129: @alexjpwalker 

```suggestion
            if file.path not in names:
```

common/java_deps.bzl
```diff
@@ -110,18 +110,24 @@ def _java_deps_impl(ctx):
...
             if ctx.attr.maven_name and file.path not in mapping:
                 fail("{} does not have associated Maven coordinate".format(file.owner))
             filename = mapping.get(file.path, file.basename).replace('.', '-').replace(':', '-')
             if filename in filenames:
                 print("Excluded duplicate: {}".format(filename))
                 continue # do not pack JARs with same name
-            names[file.path] = ctx.attr.java_deps_root + filename + ".jar"
+            for jarPattern in outputPathOverrides:
+                if file.basename == jarPattern or (jarPattern.endswith("*") and file.basename.startswith(jarPattern.rstrip("*"))):
+                    names[file.path] = outputPathOverrides.get(jarPattern) + filename + ".jar"
```
vmax, line 127: @alexjpwalker `jarPattern` is guaranteed to exist:

```suggestion
                    names[file.path] = outputPathOverrides[jarPattern] + filename + ".jar"
```

common/java_deps.bzl
```diff
@@ -110,18 +110,24 @@ def _java_deps_impl(ctx):
...
+                    continue
```
vmax, line 128: @alexjpwalker as soon as it matched, I believe we should `break`; not `continue`

alexjpwalker: oops

grabl (approved):

vmax (approved):

PR #268, opened 2020-11-11
Parse maven coordinates from 4/5 elements

hatstand:

Fix parsing of maven dependencies that do not conform to the `groupId:artifactId:version` standard, e.g., https://github.com/grpc/grpc-java/blob/bf191cb5ea56fe96e0ffa4c8936d3096c738fb07/repositories.bzl#L18

Adds more forgiving parsing of maven coordinates by using the first two values as the group & artifact IDs and the last value as the version, ignoring any other values inbetween.

### DIFF
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -19,7 +19,11 @@
 
 
 def _parse_maven_coordinates(coordinates_string, enforce_version_template=True):
-    group_id, artifact_id, version = coordinates_string.split(":")
+    coordinates = coordinates_string.split(':')
+    # Maven coordinates in the bazel ecosystem can include more than three fields.
+    # The group and artifact IDs are always the first 2 and the version is always the last field.
+    group_id, artifact_id = coordinates[0:2]
+    version = coordinates[-1]
     if enforce_version_template and version != "{pom_version}":
         fail("should assign {pom_version} as Maven version via `tags` attribute")
     return struct(
@@ -27,7 +31,6 @@ def _parse_maven_coordinates(coordinates_string, enforce_version_template=True):
         artifact_id = artifact_id,
         version = version
     )
-
 
 def _generate_version_file(ctx):
     version_file = ctx.file.version_file


lolski:
Apologies for only getting to this PR now!

Can you outline the reasoning behind the PR? From the first glance I'm not sure if relaxing the version requirement like this is a good idea. Shouldn't we try to adhere to the Maven standard as much as possible?

hatstand:
It's not entirely clear from the [maven docs](https://maven.apache.org/pom.html#dependencies) whether coordinates like this are really supposed to be a thing or not but as a practical matter, they exist in the bazel ecosystem as linked to in the description and they work with the `maven_install` target.

I don't really know why the linked code refers to the `jar` specifically though so maybe that's actually incorrect?

lolski:
Interersting. The instance you shared above is an example of coordinate with 4 elements whereas [this one](https://github.com/grpc/grpc-java/blob/bf191cb5ea56fe96e0ffa4c8936d3096c738fb07/repositories.bzl#L36) has 5 elements.

We're trying to maintain strictness in the codebase as much as possible, and try to not have any "loose ends". Therefore I wonder, is it really safe for us to simple ignore the middle elements like this?

hatstand:
I _think_ it's correct to ignore the other fields for this use case. Those fields mean something within bazel but not as a maven dependency, so they should be dropped before trying to publish.

lolski:
> Those fields mean something within bazel but not as a maven dependency

Do you happen to know how these elements are used within Bazel?

I'm interested to know how many elements can we expect (eg., only 4 or 5 elements, or more). And additionally, what do these elements represent.

hatstand:
Sorry, I don't really have any detailed knowledge beyond the `jar` and architecture fields observed above.

lolski:
Cool @hatstand . Would you mind adding a comment in the code of that fact that we're intentionally skipping the 3rd field onwards to the (but excluding) the last element? I think it is subtle enough to warrant such comment.

Otherwise, it looks good and ready to merge!

hatstand:
@lolski Done.

PR #272, opened 2020-12-02
allow conventional SNAPSHOT maven releases

### LINKED ISSUE #230: Can't deploy maven snapshots
This may be known/intentional or I may be completely wrong so I thought I'd write this up to check and potentially as ad-hoc documentation?

I've been looking at publishing snapshots to `oss.sonatype.org`.

I came across a few issues:
1. `deploy.py` wants a (git) sha for the version. maven/Sonatype want a `-SNAPSHOT` suffix.
1. The maven 3 `SNAPSHOT` model wants automatic versioning (changing `SNAPSHOT` to timestamp in filenames) and the writing of `maven-metadata.xml` to record this.

It is possible to disable the check in `deploy.py` and push a `SNAPSHOT` suffix without the extra behavior above. Documentation seems to indicate maven 3 will never write this format. Not sure if the various clients will read it (and it has a lot of other negative behaviors related to caching anyway ...)

Do I have this right? Anything I'm missing?

Thanks!

SrodriguezO:

Allow snapshot releases with the standard -SNAPSHOT naming convention.

Update the snapshot regex. I've verified the change.

Closes #230

### DIFF
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -111,7 +111,7 @@ if version is None or len(version.text) == 0:
 version = version.text
 
 snapshot = 'snapshot'
-version_snapshot_regex = '^[0-9|a-f|A-F]{40}$'
+version_snapshot_regex = '^[0-9|a-f|A-F]{40}$|.*-SNAPSHOT$'
 release = 'release'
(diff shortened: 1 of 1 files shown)


lolski:
Thank you @SrodriguezO !

PR #278, opened 2021-01-04
Update grabl automation yaml syntax

lriuui0x0:

Recently we added the support for background `monitor` script. Therefore we change the foreground job script from `script` to `command`.


vmax:
Superseded by https://github.com/graknlabs/bazel-distribution/commit/9ee5a014721343e7ce0c01056e9326b3ee3862b1

PR #290, opened 2021-03-17
Fix deploy_brew rule for non-zip targets

vmax:

Fix deploying non-ZIP (e.g. `.dmg` as in Grakn Workbase) targets to Homebrew repositories

Detect extension of the file and only validate ZIP correctness if it's in fact a ZIP archive.

### DIFF
--- a/brew/templates/deploy.py
+++ b/brew/templates/deploy.py
@@ -98,18 +98,18 @@ try:
     distribution_url = get_distribution_url_from_formula(formula_content)
     print('Attempting to match the checksums of local distribution and Github distribution from "{}"...'.format(distribution_url))
     _, ext = os.path.splitext(distribution_url)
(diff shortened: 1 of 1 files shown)


brew/templates/deploy.py
```diff
@@ -97,16 +97,19 @@ def verify_zip_file(fn):
     formula_content = formula_template.replace('{version}', version).replace('{sha256}', checksum_of_distribution_local)
     distribution_url = get_distribution_url_from_formula(formula_content)
     print('Attempting to match the checksums of local distribution and Github distribution from "{}"...'.format(distribution_url))
+    _, ext = os.path.splitext(distribution_url)
+    fn = 'distribution-github' + ext
```
lolski, line 101: can we call it `filename` rather than `fn`

PR #292, opened 2021-03-21
Improve assemble_maven to support merging JARs

vmax:

Allow merging JARs without Maven coordinates into a JAR being assembled by `assemble_maven`

Merge dependant non-Maven JARs into the deployed JAR


lolski (approved):
Approved. Let's proceed with refactoring the rule to simplify the steps even more.

PR #299, opened 2021-04-07
Allow overwriting existing image when deploying packer image

lriuui0x0:

Allow overwriting existing image when deploying packer image.

Adding overwrite attribute for packer deploy rule.

### DIFF
--- a/packer/rules.bzl
+++ b/packer/rules.bzl
@@ -70,7 +70,7 @@ deploy_packer = rule(
         "overwrite": attr.bool(
             mandatory = False,
             default = False,
-            doc = "If overwriting the image if the same named image already exists",
+            doc = "Overwrite already-existing image",
         ),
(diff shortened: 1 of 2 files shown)


packer/rules.bzl
```diff
@@ -66,6 +67,11 @@ deploy_packer = rule(
             allow_single_file = [".packer.tar"],
             doc = "`assemble_packer` label to be deployed.",
         ),
+        "overwrite": attr.bool(
+            mandatory = False,
+            default = False,
+            doc = "If overwriting the image if the same named image already exists",
```
vmax, line 73: @lriuui0x0 could be phrased simpler:

```suggestion
            doc = "Overwrite already-existing image",
```

vmax (approved):
@lriuui0x0 please state the PR goal better: more "why" we need it, not "what" we've done

PR #300, opened 2021-05-02
Fix JarAssembler to account for already-present prefix

### LINKED ISSUE #298: Source JARs produced by assemble-maven have inconsistent package prefix
## Description

Source JARs produced by assemble-maven have inconsistent package prefix. The inconsistency was observed between `protocol` and `grabl-tracing`.

Source JAR is produced with the prefix appended twice `grakn/protocol/grakn/protocol`.

## Environment

1. OS (where Grakn server runs): The issue was observed on a Mac

## Reproducible Steps

Steps to create the smallest reproducible scenario:

Go to `protocol` and build a Maven artifact there:
1. `git checkout 71b277adb6b2ad249da4d1f5e2ae9079b0d70a48`
2. `bazel build //grpc/java:assemble-maven`
3. `vim bazel-bin/grpc/java/io.grakn.protocol:grakn-protocol-sources.jar`

Next, go to `grabl-tracing and build a Maven artifact there too:
1. `git checkout e0ca77dc22585a886e4d831c6d6384f162d891cd`
2. `bazel build //client:assemble-maven`
3. `vim bazel-bin/client/io.grabl:grabl-tracing-client-sources.jar`

## Expected Output

Source JAR is produced with the correct package prefix in both `protocol` and `grabl-tracing`:
- In the case of `protocol`, an entry in the source JAR should be prefixed with `grakn/protocol`, eg., `grakn/protocol/AnswerProto.java`.
- In the case of `grabl-tracing`, an entry in the source JAR should be prefixed with `grabl/tracing`, eg., `grabl/tracing/client/GrablTokenAuthClientInterceptor.java`

## Actual Output

Source JAR is produced with the correct package prefix  for `grabl-tracing`, but not `protocol`:
- In the case of `protocol`, an entry in the source JAR is prefixed with `grakn/protocol`twice, ie., `grakn/protocol/grakn/protocol/AnswerProto.java`.
- In the case of `grabl-tracing`, an entry in the source JAR is correctly prefixed with `grabl/tracing`, eg., `grabl/tracing/client/GrablTokenAuthClientInterceptor.java`

vmax:

Allow assembling source JARs where sources or them come from mixed sources (both generated and hand-written).

Only prepend `source_jar_prefix` if it's not already present; this way, for Grabl Tracing library these changes are made (with `source_jar_prefix` = `grabl/tracing/`):

`protocol/util/ProtobufUUIDUtil.java` => `grabl/tracing/protocol/util/ProtobufUUIDUtil.java` (prefix added as it was not there before)
`grabl/tracing/protocol/TracingProto.java` => `grabl/tracing/protocol/TracingProto.java` (not modified as it already contains the correct prefix)

Fix #298

### DIFF
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -54,7 +54,11 @@ class JarAssembler : Callable<Unit> {
                         }
                         entryNames.add(entry.name)
                         BufferedInputStream(jarZip.getInputStream(entry)).use { inputStream ->
-                            val newEntry = ZipEntry(prefix + entry.name)
+                            var name = entry.name
+                            if (!name.startsWith(prefix)) {
+                                name = prefix + name;
+                            }
+                            val newEntry = ZipEntry(name)
                             out.putNextEntry(newEntry)
                             inputStream.copyTo(out, 1024)
                         }


vmax:
Closing as after discussion w @lolski we agreed to take on a better approach (outlined in https://github.com/vaticle/bazel-distribution/issues/298#issuecomment-810976712)

PR #305, opened 2021-05-05
Fix assemble_maven to produce valid source JARs

### LINKED ISSUE #298: Source JARs produced by assemble-maven have inconsistent package prefix

vmax:

Allow assembling source JARs where sources or them come from mixed sources (both generated and hand-written).

Modify `JarAssembler` to place files inside of source JAR according to the `package` statement in the source file. This allows us to produce always-correct source JARs as well as eliminates the need for passing `source_jar_prefix`

Fix #298

### DIFF
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -5,6 +5,11 @@ import java.io.BufferedInputStream
 import java.io.BufferedOutputStream
 import java.io.File
 import java.io.FileOutputStream
+import java.lang.RuntimeException
+import java.nio.ByteBuffer
+import java.nio.charset.Charset
+import java.nio.file.Path
+import java.nio.file.Paths
 import java.util.*
 import java.util.concurrent.Callable
 import java.util.zip.ZipEntry
@@ -15,12 +20,10 @@ import kotlin.system.exitProcess
 
 @Command(name = "jar-assembler", mixinStandardHelpOptions = true)
 class JarAssembler : Callable<Unit> {
+    val javaPackageRegex = Regex("package\\s+([a-zA_Z_][\\.\\w]*);")
 
     @Option(names = ["--output"], required = true)
     lateinit var output_file: File
-
-    @Option(names = ["--prefix"])
-    lateinit var prefix: String
 
     @Option(names = ["--group-id"])
     var groupId = ""
@@ -34,32 +37,61 @@ class JarAssembler : Callable<Unit> {
     @Option(names = ["--jars"], split = ";")
     lateinit var jars: Array<File>
 
-    val entryNames = mutableSetOf<String>()
+    private val entryNames = mutableSetOf<String>()
+    val entries = HashMap<String, ByteArray>()
+
+    /**
+     * For path "a/b/c.java" inserts "a/" and "a/b/ into `entries`
+     */
+    private fun addDirectories(path: Path) {
+        for (i in path.nameCount-1 downTo 1) {
+            val subPath = path.subpath(0, i).toString() + "/"
+            entries[subPath] = ByteArray(0)
+        }
+    }
 
     override fun call() {
         ZipOutputStream(BufferedOutputStream(FileOutputStream(output_file))).use { out ->
             if (pomFile != null) {
-                val pomFileEntry = ZipEntry("META-INF/maven/${groupId}/${artifactId}/pom.xml")
-                out.putNextEntry(pomFileEntry)
-                out.write(pomFile!!.readBytes())
+                val pomPath = "META-INF/maven/${groupId}/${artifactId}/pom.xml"
+                entries[pomPath] = pomFile!!.readBytes()
+                addDirectories(Paths.get(pomPath))
             }
             for (jar in jars) {
                 ZipFile(jar).use { jarZip ->
                     jarZip.entries().asSequence().forEach { entry ->
                         if (entry.name.contains("META-INF")) {
+                            // pom.xml will be added by us
                             return@forEach
                         }
                         if (entryNames.contains(entry.name)) {
+                            throw RuntimeException("duplicate entry in the JAR: ${entry.name}")
+                        }
+                        if (entry.isDirectory) {
+                            // needed directories would be added by us
                             return@forEach
                         }
                         entryNames.add(entry.name)
                         BufferedInputStream(jarZip.getInputStream(entry)).use { inputStream ->
-                            val newEntry = ZipEntry(prefix + entry.name)
-                            out.putNextEntry(newEntry)
-                            inputStream.copyTo(out, 1024)
+                            val sourceFileBytes = inputStream.readBytes()
+                            val sourceFile = sourceFileBytes.toString(Charset.forName("UTF-8"))
+                            var resultLocation = entry.name
+                            // files in source JARs are moved according to their `package` statement
+                            if (entry.name.endsWith(".java")) {
+                                val sourceFileName = Paths.get(entry.name).fileName.toString()
+                                val sourceFilePackage = javaPackageRegex.find(sourceFile)?.groups?.get(1)?.value ?: throw RuntimeException("could not obtain package of ${entry.name}")
+                                resultLocation = "${sourceFilePackage.replace(".", "/")}/$sourceFileName"
+                            }
+                            entries[resultLocation] = sourceFileBytes
+                            addDirectories(Paths.get(resultLocation))
                         }
                     }
                 }
+            }
+            entries.keys.sorted().forEach {
+                val newEntry = ZipEntry(it)
+                out.putNextEntry(newEntry)
+                out.write(entries[it]!!)
             }
         }
     }
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 ## assemble_maven
 
 <pre>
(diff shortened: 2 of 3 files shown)


maven/JarAssembler.kt
```diff
@@ -34,35 +37,64 @@ class JarAssembler : Callable<Unit> {
...
     lateinit var jars: Array<File>
 
-    val entryNames = mutableSetOf<String>()
+    private val entryNames = mutableSetOf<String>()
+    val entries = HashMap<String, ByteArray>()
+
+    /**
+     * For path "a/b/c.java" inserts "a/" and "a/b/ into `entries`
+     */
+    private fun addDirectories(path: Path) {
```
lolski, line 46: let's call it `preCreateDirectories`?

maven/JarAssembler.kt
```diff
@@ -34,35 +37,64 @@ class JarAssembler : Callable<Unit> {
...
+                        if (entry.isDirectory) {
+                            // needed directories would be added by us
                             return@forEach
                         }
                         entryNames.add(entry.name)
                         BufferedInputStream(jarZip.getInputStream(entry)).use { inputStream ->
-                            val newEntry = ZipEntry(prefix + entry.name)
-                            out.putNextEntry(newEntry)
-                            inputStream.copyTo(out, 1024)
+                            val sourceFileBytes = inputStream.readBytes()
```
lolski, line 76: can we extract this bit into a function that is named accordingly:
```
                            val sourceFileBytes = inputStream.readBytes()
                            val sourceFile = sourceFileBytes.toString(Charset.forName("UTF-8"))
                            var resultLocation = entry.name
                            // files in source JARs are moved according to their `package` statement
                            if (entry.name.endsWith(".java")) {
                                val sourceFileName = Paths.get(entry.name).fileName.toString()
                                val sourceFilePackage = javaPackageRegex.find(sourceFile)?.groups?.get(1)?.value ?: throw RuntimeException("could not obtain package of ${entry.name}")
                                resultLocation = "${sourceFilePackage.replace(".", "/")}/$sourceFileName"
                            }
                            entries[resultLocation] = sourceFileBytes
                            addDirectories(Paths.get(resultLocation))
```

maven/JarAssembler.kt
vmax: @lolski addressed in https://github.com/vaticle/bazel-distribution/pull/305/commits/a843c9a2834b9693197d6e45ec799274a61284b2

maven/JarAssembler.kt
vmax: @lolski addressed in https://github.com/vaticle/bazel-distribution/pull/305/commits/a843c9a2834b9693197d6e45ec799274a61284b2

maven/JarAssembler.kt
```diff
@@ -34,35 +36,73 @@ class JarAssembler : Callable<Unit> {
...
+    }
 
     override fun call() {
         ZipOutputStream(BufferedOutputStream(FileOutputStream(output_file))).use { out ->
             if (pomFile != null) {
-                val pomFileEntry = ZipEntry("META-INF/maven/${groupId}/${artifactId}/pom.xml")
-                out.putNextEntry(pomFileEntry)
-                out.write(pomFile!!.readBytes())
+                val pomPath = "META-INF/maven/${groupId}/${artifactId}/pom.xml"
+                entries[pomPath] = pomFile!!.readBytes()
```
lolski, line 75: Can we make `preCreateDirectories` _side-effect-free_ and returns created entries? Then, you would merge those entries with the global `entries` variable:

```
                entries += preCreateDirectories(Paths.get(pomPath))
                entries[pomPath] = pomFile!!.readBytes()
```

lolski: Also I think it's nicer if you do the pre-create before rather than after.

maven/JarAssembler.kt
```diff
@@ -34,35 +36,73 @@ class JarAssembler : Callable<Unit> {
...
+                        if (entry.isDirectory) {
+                            // needed directories would be added by us
                             return@forEach
                         }
                         entryNames.add(entry.name)
                         BufferedInputStream(jarZip.getInputStream(entry)).use { inputStream ->
-                            val newEntry = ZipEntry(prefix + entry.name)
-                            out.putNextEntry(newEntry)
-                            inputStream.copyTo(out, 1024)
+                            prepareEntry(inputStream, entry)
```
lolski, line 94: I think the name `prepareEntry` isn't very meaningful and too generic. If I see a step "prepare", I would want to see another step "execute" and "cleanup", for example. That way, I understand that the flow of the program follows the three-stage approach of "prepare", "execute" and "cleanup".

In our program here, we only have "prepare". Therefore it actually makes the reader wondering.

Therefore I would suggest inlining the content here:

```
BufferedInputSttream(...).use { inputStream ->
        val sourceFileBytes = inputStream.readBytes()
        val resultLocation = getFinalPath(entry, sourceFileBytes)
        entries[resultLocation] = sourceFileBytes
        preCreateDirectories(Paths.get(resultLocation))
}
```

which is already clean enough since it only has 4 lines

vmax: @lolski addressed in https://github.com/vaticle/bazel-distribution/pull/305/commits/cad87304cd7785647fcfc952b1fd3152cf396490

maven/JarAssembler.kt
vmax: @lolski addressed in https://github.com/vaticle/bazel-distribution/pull/305/commits/cad87304cd7785647fcfc952b1fd3152cf396490

maven/JarAssembler.kt
```diff
@@ -34,35 +37,71 @@ class JarAssembler : Callable<Unit> {
...
+        }
+        return newEntries
+    }
+
+    private fun getFinalPath(entry: ZipEntry, sourceFileBytes: ByteArray): String {
+        return if (entry.name.endsWith(".java")) {
+            // files in source JARs are moved according to their `package` statement
+            val sourceFile = sourceFileBytes.toString(Charset.forName("UTF-8"))
+            val sourceFileName = Paths.get(entry.name).fileName.toString()
+            val sourceFilePackage = javaPackageRegex.find(sourceFile)?.groups?.get(1)?.value ?: throw RuntimeException("could not obtain package of ${entry.name}")
```
lolski, line 60:
```
private fun toStringUTF8(sourceFileBytes: ByteArray) {
  return sourceFileBytes.toString(Charset.forName("UTF-8"))
}

private fun getJavaPackage(sourceFile: String) {
  val javaPackageRegex = Regex("package\\s+([a-zA_Z_][\\.\\w]*);")
  return javaPackageRegex.find(sourceFile)?.groups?.get(1)?.value ?: throw RuntimeException("could not obtain package of ${entry.name}")
}

val sourceFile = toStringUTF8(sourceFileBytes)
val sourceFilePackage = getJavaPackage(sourceFileBytes)
```

maven/JarAssembler.kt
```diff
@@ -34,35 +36,80 @@ class JarAssembler : Callable<Unit> {
...
                         }
                         entryNames.add(entry.name)
                         BufferedInputStream(jarZip.getInputStream(entry)).use { inputStream ->
-                            val newEntry = ZipEntry(prefix + entry.name)
-                            out.putNextEntry(newEntry)
-                            inputStream.copyTo(out, 1024)
+                            val sourceFileBytes = inputStream.readBytes()
+                            val resultLocation = getFinalPath(entry, sourceFileBytes)
+                            entries[resultLocation] = sourceFileBytes
+                            preCreateDirectories(Paths.get(resultLocation))
```
lolski, line 101: Are we sure we don't need to do `entries += preCreateDirectories(Paths.get(resultLocation))`?

vmax: @lolski good point; addressed in https://github.com/vaticle/bazel-distribution/pull/305/commits/ea631b9c791ec6fc11151abc9f58d5f9795408d5

lolski (approved):

PR #307, opened 2021-05-11
Refactor the common package

lolski:

We've completely refactored the `//common` package, creating a subpackage for each rule for better code organisation.

- Split `//common` into multiple subpackages


common/rules.bzl
```diff
@@ -127,7 +127,7 @@ def _assemble_archive_prefix_file_impl(ctx):
     )
 
 
-_assemble_zip_prefix_file = rule(
+assemble_zip_prefix_file = rule(
```
lolski, line 130: Make `assemble_zip_prefix_file` public for access by `assemble_zip`

lolski:
The PR was originally intended to fix the implementation of `assemble_targz` which will handle long file path incorrectly.

However, I have reverted the fix after a discussion with @vmax. He will fix it on a follow-up PR since he found a better approach.

Hence, this PR contains just the refactors.

PR #309, opened 2021-05-13
Refactor assemble_zip rule to be consistent with assemble_targz

vmax:

Ensure behaviour of `assemble_zip` is consistent with `assemble_targz` by relying on it to do the actual work.

Fix #306 

* Rely on `assemble_targz` for `assemble_zip` implementation
* Modify `tgz2zip` to not modify the archive structure and just do the repacking

### DIFF
--- a/common/zip/rules.bzl
+++ b/common/zip/rules.bzl
@@ -15,7 +15,6 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@vaticle_bazel_distribution//common/targz:rules.bzl", "assemble_targz")
 load("@vaticle_bazel_distribution//common/tgz2zip:rules.bzl", "tgz2zip")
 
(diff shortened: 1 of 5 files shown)


lolski (approved):

common/zip/rules.bzl
```diff
@@ -16,6 +16,7 @@
 #
 
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
+load("@vaticle_bazel_distribution//common/targz:rules.bzl", "assemble_targz")
 load("@vaticle_bazel_distribution//common/tgz2zip:rules.bzl", "tgz2zip")
 
 def _assemble_archive_prefix_file_impl(ctx):
```
lolski, line 22: can we remove `assemble_zip_prefix_file` then?

lolski: what other things can be cleaned up?

vmax: @lolski yup, it's no longer needed; addressed in https://github.com/vaticle/bazel-distribution/pull/309/commits/b717657fd6f7f52e5a7dace6c0ef92e5cf637a72

vmax: @lolski `pkg_tar` is no longer used as well; addressed in https://github.com/vaticle/bazel-distribution/pull/309/commits/ec84bf3ce877608456d8e394ee070143f9edbc85

PR #310, opened 2021-05-21
Validate environment variables for deploy_brew

vmax:

Allow `deploy_brew` to fail earlier with a better error message

Prevalidate the existence of environment variable that `deploy_brew` expects


vmax:
Superseded by #313

PR #326, opened 2021-12-03
Platform JVM assembler fixes

alexjpwalker:

The `assemble_jvm_platform` rule was failing to compile due to some broken changes that went in, so we fixed them.

- Undelete config_setting: apple-code-sign (I thought it was unused but it is actually used in typedb-studio)
- Correctly unpackage the JDK on Mac/Linux using `tar`, not `jar`

### DIFF
--- a/platform/jvm/BUILD
+++ b/platform/jvm/BUILD
@@ -46,3 +46,10 @@ java_binary(
     main_class = "com.vaticle.bazel.distribution.platform.jvm.MainKt",
     visibility = ["//visibility:public"],
 )
+
+config_setting(
+    name = "apple-code-sign",
+    define_values = {
+        "APPLE_CODE_SIGN": "yes"
+    }
+)
--- a/platform/jvm/JVMPlatformAssembler.kt
+++ b/platform/jvm/JVMPlatformAssembler.kt
@@ -33,6 +33,7 @@ import com.vaticle.bazel.distribution.platform.jvm.Logging.Logger
 import com.vaticle.bazel.distribution.platform.jvm.Shell.Programs.JAR
 import com.vaticle.bazel.distribution.platform.jvm.Shell.Programs.JPACKAGE
 import com.vaticle.bazel.distribution.platform.jvm.Shell.Programs.JPACKAGE_EXE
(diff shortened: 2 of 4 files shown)


platform/jvm/BUILD
```diff
@@ -46,3 +46,10 @@ java_binary(
     main_class = "com.vaticle.bazel.distribution.platform.jvm.MainKt",
     visibility = ["//visibility:public"],
 )
+
+config_setting(
```
alexjpwalker, line 50: During the review of the previous PR I didn't see this being used anywhere, but it turns out TypeDB Studio uses it.

platform/jvm/JVMPlatformAssembler.kt
```diff
@@ -92,7 +93,10 @@ object JVMPlatformAssembler {
 
         fun extractJDK() {
             createDirectory(Path.of(JDK))
-            shell.execute(command = listOf(JAR, "xf", Path.of("..", options.jdkPath).toString()), baseDir = Path.of(JDK))
+            when (currentOS) {
+                MAC, LINUX -> shell.execute(listOf(TAR, "-xf", options.jdkPath, "-C", JDK))
```
alexjpwalker, line 97: During the previous PR review I assumed this was just a piece of technical debt that hadn't been refactored out, since this is the only place we use TAR. But it turns out that we actually strictly need TAR here, because the JDK is packaged as a TAR archive on Mac and Linux, so JAR won't work.

platform/jvm/rules.bzl
```diff
@@ -172,7 +172,7 @@ windowsWiXToolsetPath: {}
     )
 
     config_path_arg = "--config_path={}".format(config_file.path)
-    if "APPLE_CODE_SIGN" in ctx.var and ctx.var["APPLE_CODE_SIGN"].lower() in "yes, true":
+    if "APPLE_CODE_SIGN" in ctx.var and ctx.var["APPLE_CODE_SIGN"] == "yes":
```
alexjpwalker, line 175: Given that we do need the `config_setting` where `APPLE_CODE_SIGN` is specifically set to "yes", we also need to tighten up this check to align with it.

lolski (approved):

PR #330, opened 2022-01-05
Make deploy_npm take in a pre-existing auth token

alexjpwalker:

We made `deploy_npm` take in a pre-existing auth token instead of executing `npm adduser` and signing in with username/password on every run.

### Changeset

- Don't run `npm adduser` in `deploy_npm` anymore, but rather, create a `.npmrc` file (in the Bazel sandbox) from a provided environment variable `DEPLOY_NPM_TOKEN` before running `npm publish` (eliminates need for `DEPLOY_NPM_USERNAME`, `DEPLOY_NPM_PASSWORD` and `DEPLOY_NPM_EMAIL`)
- Rewrite `deploy_npm` in Kotlin
- Remove vmax from CODEOWNERS

### Explanation

- **Running `npm adduser` on every run was not the best thing to do**
This command, misleadingly, does not "add a user", but rather, it uses provided login credentials in order to generate an auth token that authorises you to publish an `npm` package. This auth token would be used once and then consigned to history, which was wasteful.

- **Now `npmjs.com` won't let us do it anymore thanks to 2FA**
Recently `npm`'s central repository appears to have added a new requirement when logging in (e.g: via `npm adduser`): it sends a OTP (one-time passcode) to your registered email address which you must provide to complete the login. This is especially awkward for automation.

- **The user must now generate their own auth token and provide it as an env var**
For each `npm` **repository** (NOT each individual package!), and for each user account that is to be used to publish artifacts - typically just a single user per repo - the user is now required to generate their own **reusable auth token**. This auth token should be securely stored (say, in an environment variable) and passed in CI to the `deploy_npm` rule as an environment variable named `DEPLOY_NPM_TOKEN`.

- **We no longer need username, password and email**
`DEPLOY_NPM_USERNAME`, `DEPLOY_NPM_PASSWORD` and `DEPLOY_NPM_EMAIL` are redundant as we are no longer signing in.

#### `typedb-protocol/grabl/automation.yml > deploy-npm-snapshot`
```sh
export DEPLOY_NPM_TOKEN=$REPO_VATICLE_NPM_TOKEN
bazel run --define version=$(git rev-parse HEAD) //grpc/nodejs:deploy-npm -- snapshot
```
#### `typedb-protocol/grabl/automation.yml > deploy-npm-release`
```sh
export DEPLOY_NPM_TOKEN=$REPO_NPM_TOKEN
bazel run --define version=$(cat VERSION) //grpc/nodejs:deploy-npm -- release
```


npm/templates/deploy.py
```diff
@@ -66,19 +67,11 @@
...
-    expect_input_file.write(expect_input_tmpl.format(
-        registry=npm_registry,
-        username=npm_username,
-        email=npm_email,
-    ))
+if repo_type == "release" and not npm_token:
+    print(
+        'token should be passed via '
+        '$DEPLOY_NPM_TOKEN env variable when deploying to a release repo'
+    )
```
alexjpwalker, line 74: `DEPLOY_NPM_TOKEN` must now be provided when deploying to a release repo (note: this is just a sanity check. Release repos typically require authentication and snapshot repos typically don't)

npm/templates/deploy.py
```diff
@@ -89,13 +82,9 @@
...
-with open(expect_input_file.name) as expect_input:
-    subprocess.check_call([
-        '/usr/bin/expect',
-    ], stdin=expect_input, env={
-        'PATH': node_path,
-        'PASSWORD': npm_password
-    })
+f = open(".npmrc", "w")
+f.write("//%s:_authToken=%s" % (npm_registry, npm_token))
+f.close()
```
alexjpwalker, line 87: Previously we relied on `npm adduser` to generate this `.npmrc` file by signing in to the remote npm repo. Now, we create it ourselves manually based on the remote repo name and the provided auth token.

npm/templates/deploy.py
```diff
@@ -42,10 +42,11 @@
 }
 npm_registry = npm_repositories[repo_type]
 
-npm_username, npm_password, npm_email = (
+npm_username, npm_password, npm_email, npm_token = (
     os.getenv('DEPLOY_NPM_USERNAME'),
```
lolski, line 46: Do we need to take in `DEPLOY_NPM_USERNAME` and `DEPLOY_NPM_PASSWORD` still?

npm/templates/deploy.py
lolski: Are we sure we don't have to provide authentication credential when deploying to the snapshot repo? Doesn't that mean anyone can deploy a snapshot artefact if they know the URL?

npm/templates/deploy.py
```diff
@@ -42,10 +42,11 @@
 }
 npm_registry = npm_repositories[repo_type]
 
-npm_username, npm_password, npm_email = (
+npm_username, npm_password, npm_email, npm_token = (
```
lolski, line 45: I didn't mention it in our discussion, but we should rewrite `deploy.py` in Kotlin. Remember that we're phasing out Python from our build systems.

npm/templates/deploy.py
alexjpwalker: It's a sufficient (and not necessary) condition to provide $DEPLOY_NPM_TOKEN to a snapshot repo. My thought was that some packages in snapshot repos might not require authentication to publish.

But to be honest I've never seen an example of this, so I'll update the check to make DEPLOY_NPM_TOKEN always mandatory.

npm/templates/deploy.py
alexjpwalker: That's a fair point, token authentication doesn't require username/password, I'll delete them

common/Logging.kt
```diff
@@ -0,0 +1,35 @@
+/*
```
alexjpwalker, line 1: Moved from `platform.jvm` to `common` as it is now used by both `assemble_jvm_platform` and `deploy_npm`

It's depended on by `Shell`.

common/shell/BUILD
```diff
@@ -0,0 +1,31 @@
+#
```
alexjpwalker, line 1: Moved from `platform.jvm` to `common` as it is now used by both `assemble_jvm_platform` and `deploy_npm`

common/shell/Shell.kt
```diff
@@ -1,13 +1,13 @@
-package com.vaticle.bazel.distribution.platform.jvm
+package com.vaticle.bazel.distribution.common.shell
```
alexjpwalker, line 1: Moved from `platform.jvm` to `common` as it is now used by both `assemble_jvm_platform` and `deploy_npm`

The intent is for this class to be the standard for running shell commands in `bazel-distribution`.

npm/BUILD
```diff
@@ -32,3 +33,23 @@ py_binary(
     srcs = ["assemble.py"],
     visibility = ["//visibility:public"],
 )
+
+kt_jvm_library(
```
alexjpwalker, line 37: The Kotlin library and backing Java binary that will now power `deploy_npm`.

npm/DeployNPM.kt
```diff
@@ -0,0 +1,36 @@
+/*
```
alexjpwalker, line 1: Entry point for `deploy_npm`

npm/Deployer.kt
```diff
@@ -0,0 +1,132 @@
+/*
```
alexjpwalker, line 1: Core logic of `deploy_npm`

npm/Deployer.kt
```diff
@@ -0,0 +1,132 @@
...
+import com.vaticle.bazel.distribution.npm.Deployer.CommandLineParams.Keys.SNAPSHOT_REPO
+import com.vaticle.bazel.distribution.npm.Deployer.Env.DEPLOY_NPM_TOKEN
+import com.vaticle.bazel.distribution.npm.Deployer.Env.PATH
+import com.vaticle.bazel.distribution.npm.Deployer.Options.RepositoryType.*
+import picocli.CommandLine
+import java.nio.file.Files
+import java.nio.file.Path
+
+class Deployer(options: Options) {
+    private val logger = Logger(logLevel = DEBUG)
```
alexjpwalker, line 41: For a `bazel run` target it's helpful to use the DEBUG log level as it tells you whether each subcommand succeeded. We typically use ERROR for `bazel build` targets.

npm/Deployer.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    fun deploy() {
+        configureAuthToken()
+        publishPackage()
+    }
+
+    private fun configureAuthToken() {
+        val npmToken = System.getenv(DEPLOY_NPM_TOKEN)
+            ?: throw IllegalArgumentException("token should be passed via \$$DEPLOY_NPM_TOKEN env variable")
+        Files.writeString(Path.of(".npmrc"), "//$registryURL:_authToken=$npmToken")
```
alexjpwalker, line 53: `npm publish` requires token-based authentication provided by a `.npmrc` file.

It will search for `.npmrc` in the directory it's run from first. This is helpful to us as it allows us to store `.npmrc` in the Bazel sandbox, rather than `$HOME/.npmrc` (the default location)

Previously we generated this `.npmrc` by running `npm adduser` and providing username/password, but this was wasteful - there's no need to generate a new auth token and log in every run. `npmjs` central registry recently added a new layer of security, where you are challenged for a OTP (one-time password) sent to your email. This makes it very difficult to automate - and furthermore, the new reusable token-based solution is just better.

I've included instructions on generating a reusable auth token in `npm/README.md`.

npm/Deployer.kt
```diff
@@ -0,0 +1,132 @@
...
+            command = listOf("npm", "publish", "--registry=$registryURL", "deploy_npm.tgz"),
+            env = mapOf(PATH to pathEnv()))
+    }
+
+    private fun pathEnv(): String {
+        val commonPaths = listOf(realPath(npmPath).parent)
+        val otherPaths = when (currentOS) {
+            MAC, LINUX -> listOf("/usr/bin", "/bin/")
+            else -> listOf()
+        }
```
alexjpwalker, line 67: On Mac and Linux, we need `/usr/bin` and `/bin` on the PATH to ensure we can find `bash`

npm/Deployer.kt
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private fun publishPackage() {
+        Shell(logger = logger, verbose = true).execute(
+            command = listOf("npm", "publish", "--registry=$registryURL", "deploy_npm.tgz"),
+            env = mapOf(PATH to pathEnv()))
+    }
+
+    private fun pathEnv(): String {
+        val commonPaths = listOf(realPath(npmPath).parent)
```
alexjpwalker, line 63: It's worth noting (because this tripped me up!) - that providing `npmPath.parent` raw, without calling `realPath` to follow symlinks, results in an error running `npm`: `node not found`.

This is because `npmPath` is a symlink to the `npm` executable. So, although providing its directory (before following the symlink) is sufficient to find `npm`, it's **not** sufficient to find `node` - which lives in the same directory as `npm` **after** following the symlink.

npm/Deployer.kt
```diff
@@ -0,0 +1,132 @@
...
+        }
+        return (commonPaths + otherPaths).joinToString(":")
+    }
+
+    private fun realPath(path: String): Path {
+        val pathObj = Path.of(path)
+        return when (Files.exists(pathObj)) {
+            true -> pathObj.toRealPath()
+            false -> pathObj
+        }
```
alexjpwalker, line 76: `toRealPath` actually throws if the file doesn't exist, so we need to sanity-check the path.

npm/README.md
```diff
@@ -0,0 +1,15 @@
+# NPM rule usage guide
```
alexjpwalker, line 1: Here we explain how to generate an auth token, because it's now a requirement to use `deploy_npm`.

I don't expect these instructions to change in the next couple of years.

npm/rules.bzl
```diff
@@ -83,39 +83,37 @@ assemble_npm = rule(
 )
```
alexjpwalker, line 83: Updating `deploy_npm` rule implementation to be backed by Kotlin script instead of Python

npm/templates/deploy.sh
```diff
@@ -0,0 +1,24 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+exec java -jar "{DEPLOYER_PATH}" \
+    --npm_path="{NPM_PATH}" \
+    --snapshot_repo="{SNAPSHOT_REPO}" \
+    --release_repo="{RELEASE_REPO}" "$*"
```
alexjpwalker, line 24: Pass in templated args and, at the end, positional parameters from the command line (intended to be either `-- snapshot` or `-- release`)

platform/jvm/AppleCodeSigner.kt
```diff
@@ -1,5 +1,6 @@
 package com.vaticle.bazel.distribution.platform.jvm
```
alexjpwalker, line 1: Moved some components to `common` because they are now used by `deploy_npm`

platform/jvm/BUILD
```diff
@@ -33,6 +33,7 @@ kt_jvm_library(
     srcs = glob(["*.kt"]),
```
alexjpwalker, line 33: Moved `shell` to `common` because it is now used by `deploy_npm`

platform/jvm/JVMPlatformAssembler.kt
```diff
@@ -1,8 +1,10 @@
 package com.vaticle.bazel.distribution.platform.jvm
```
alexjpwalker, line 1: Moved some components to `common` because they are now used by `deploy_npm`

platform/jvm/Logging.kt
```diff
@@ -1,16 +0,0 @@
-package com.vaticle.bazel.distribution.platform.jvm
```
alexjpwalker, line 1: Moved to `common` because now used by `deploy_npm`

platform/jvm/MacAppNotarizer.kt
```diff
@@ -1,5 +1,6 @@
 package com.vaticle.bazel.distribution.platform.jvm
```
alexjpwalker, line 1: Moved some components to `common` because they are now used by `deploy_npm`

platform/jvm/Main.kt
```diff
@@ -27,7 +27,7 @@ fun parseCommandLine(args: Array<String>): Options {
     val commandLine = CommandLine(CommandLineParams())
     val parseResult: CommandLine.ParseResult = commandLine.parseArgs(*args)
     assert(parseResult.asCommandLineList().size == 1)
-    val parameters: CommandLineParams = parseResult.asCommandLineList()[0].getCommand<CommandLineParams>()
+    val parameters: CommandLineParams = parseResult.asCommandLineList()[0].getCommand()
```
alexjpwalker, line 30: refactor out unused symbol

platform/jvm/Options.kt
```diff
@@ -21,15 +21,15 @@
 
 package com.vaticle.bazel.distribution.platform.jvm
 
+import com.vaticle.bazel.distribution.common.Logging.Logger
+import com.vaticle.bazel.distribution.common.Logging.LogLevel
```
alexjpwalker, line 25: formatted alphabetically

platform/jvm/ShellArgs.kt
```diff
@@ -0,0 +1,19 @@
+package com.vaticle.bazel.distribution.platform.jvm
```
alexjpwalker, line 1: Shell file was moved to `common` but we want to keep these args in `platform.jvm`.

common/shell/Shell.kt
```diff
@@ -49,26 +49,10 @@ class Shell(private val verbose: Boolean = false, private val printSensitiveData
...
 
-        class Argument(val value: String, val printable: Boolean = true) {
+        class Argument(val value: String, private val printable: Boolean = true) {
             override fun toString(): String {
                 return if (printable) value else "(hidden)"
             }
         }
     }
-
-    object Programs {
```
alexjpwalker, line 59: These stayed in `jvm.platform` because they aren't used by `deploy-npm`

npm/rules.bzl
```diff
@@ -83,39 +83,37 @@ assemble_npm = rule(
...
 
 
-def _deploy_npm(ctx):
+def _deploy_npm_impl(ctx):
```
alexjpwalker, line 86: Explanation:

1. Declare that we're creating a (shell) script to deploy an NPM package
2. Create the script from the template, `templates/deploy.py`, using variable substitution
3. When `bazel run` is invoked, run the script, passing the package to be deployed, the Java JAR containing the deployer code, and `npm`, and symlink `deploy_npm.tgz` to the package being deployed for convenience

npm/templates/deploy.py
```diff
@@ -1,107 +0,0 @@
-#!/usr/bin/env python
```
alexjpwalker, line 1: Migrated from Python to Kotlin

npm/templates/deploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
alexjpwalker, line 1: Template for the bash script that will be executed when `bazel run` is called

lolski (changes requested):

README.md
```diff
@@ -271,6 +271,20 @@ deploy_npm(<a href="#deploy_npm-name">name</a>, <a href="#deploy_npm-release">re
 </pre>
```
alexjpwalker, line 271: New docs of `deploy_npm` rule

npm/BUILD
```diff
@@ -16,19 +16,3 @@
 # specific language governing permissions and limitations
```
alexjpwalker, line 16: This package no longer contains any rules since we extracted them into `assemble` and `deploy` subpackages

npm/assemble/BUILD
```diff
@@ -0,0 +1,34 @@
+#
```
alexjpwalker, line 1: Package containing `assemble_npm`

npm/assemble/rules.bzl
```diff
@@ -0,0 +1,83 @@
+#
```
alexjpwalker, line 1: File containing definition of `assemble_npm` rule only (note that this is now less cluttered than before when it had 2 rule implementations in the same file, neither of which was trivial)

npm/deploy/BUILD
```diff
@@ -0,0 +1,51 @@
+#
```
alexjpwalker, line 1: Package containing `deploy_npm` rule

npm/deploy/DeployNPM.kt
```diff
@@ -0,0 +1,26 @@
+/*
```
alexjpwalker, line 1: Entry point of `deploy-npm`

npm/deploy/Deployer.kt
```diff
@@ -0,0 +1,67 @@
+/*
```
alexjpwalker, line 1: Core logic of `deploy-npm`

npm/deploy/Deployer.kt
```diff
@@ -0,0 +1,67 @@
...
+        publishPackage()
+    }
+
+    private fun configureAuthToken() {
+        Files.writeString(Path.of(".npmrc"), "//${options.registryURL}:_authToken=${options.npmToken}")
+    }
+
+    private fun publishPackage() {
+        Shell(logger = logger, verbose = true).execute(
+            command = listOf("npm", "publish", "--registry=${options.registryURL}", "deploy_npm.tgz"),
```
alexjpwalker, line 47: Note that `deploy_npm.tgz`'s name is baked into the Skylark rule, it symlinks to the thing we want to deploy

npm/deploy/Options.kt
```diff
@@ -0,0 +1,73 @@
+package com.vaticle.bazel.distribution.npm.deploy
```
alexjpwalker, line 1: All options for `deploy-npm` rule

npm/deploy/deploy.sh.template
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
alexjpwalker, line 1: Template script that is an entry point for running the `deploy-npm` jar

npm/deploy/rules.bzl
```diff
@@ -0,0 +1,90 @@
+#
```
alexjpwalker, line 1: Contains the implementation of `deploy-npm`

npm/rules.bzl
```diff
@@ -17,121 +17,8 @@
 # under the License.
```
alexjpwalker, line 17: This file no longer contains any rule implementations itself, but rather re-exports the implementations in the subpackages, helping to keep the codebase clean / adherent to single responsibility principle.

npm/deploy/Options.kt
```diff
@@ -0,0 +1,73 @@
...
+        get() {
+            return System.getenv(DEPLOY_NPM_TOKEN)
+                ?: throw IllegalArgumentException("token should be passed via \$${DEPLOY_NPM_TOKEN} env variable")
+        }
+
+    companion object {
+        fun of(args: Array<String>): Options {
+            val commandLine = CommandLine(Options())
+            val parseResult: CommandLine.ParseResult = commandLine.parseArgs(*args)
+            assert(parseResult.asCommandLineList().size == 1)
```
lolski, line 45: why can we guarantee that there's exactly one element returned by `asCommandLineList`?

Also since there's no guarantee that `asCommandLineList()` is hermetic (it most likely is, but we can't say for sure), this is the better style:
```
val cliList = parseResult.asCommandLineList()
assert (cliList.size == 1)
return cliList[0].getCommand()
```

npm/deploy/Options.kt
```diff
@@ -0,0 +1,72 @@
...
+
+import com.vaticle.bazel.distribution.npm.deploy.Options.CommandLineParams.NPM_PATH
+import com.vaticle.bazel.distribution.npm.deploy.Options.CommandLineParams.RELEASE_REPO
+import com.vaticle.bazel.distribution.npm.deploy.Options.CommandLineParams.SNAPSHOT_REPO
+import com.vaticle.bazel.distribution.npm.deploy.Options.Env.DEPLOY_NPM_TOKEN
+import com.vaticle.bazel.distribution.npm.deploy.Options.RepositoryType.RELEASE
+import com.vaticle.bazel.distribution.npm.deploy.Options.RepositoryType.SNAPSHOT
+import picocli.CommandLine
+
+class Options {
```
alexjpwalker, line 11: Options are constructed from command line arguments and environment variables.

We use picocli for command line parsing. Argument names are defined in the inner object `Options.CommandLineParams`. Env var names are defined in `Options.Env`.

Parameters after a `--` in the command are passed in to the `params` variable (because of its annotation `@CommandLine.Parameters`). For example, in `java -jar DeployNPM.jar --npm_path /usr/bin/node -- snapshot`, `params` will have a value of `["snapshot"]`.

We enumerate, and validate, these positional parameters in `Options.RepositoryType`, ensuring that the user may only pass in a single positional parameter; this parameter represents the _repo type_; and the repo type must be "snapshot" or "release".

lolski (approved):

npm/deploy/Options.kt
alexjpwalker: See the documentation for `asCommandLineList`

> Returns this ParseResult as a list of CommandLine objects, one for each matched command/subcommand.
>    * Note that for repeatable subcommands, there may be multiple commands at each level of the hierarchy in the returned list.

alexjpwalker:
### Principles applied in PR review

#### Bazel rule implementation

- When modifying a Python implementation of a Bazel rule, rewrite it in Kotlin
- "One Bazel rule implementation per package" is a good level of encapsulation
  - We found that `rules.bzl` became cluttered when both `assemble_npm` and `deploy_npm`'s implementations lived there. Breaking them out into separate packages helps us adhere to the single responsibility principle.
- When creating a Kotlin file with a `main` method to implement a Bazel rule, name the file after the rule: for example, `deploy_npm` runs `DeployNPM.kt`
  - Although `Main.kt` feels appropriate, if we used that name, almost every file in the repo would eventually be named `Main.kt`, so we don't want that.  
- When writing usage guidelines for a Bazel rule, place them in the 'doc' field of the rule definition
- Our Bazel rule docs and their parameter docs should use full stops

#### Naming

- We will begin to name our template files `{filename}.{ext}.template` (e.g: `deploy.sh.template`), so there is no need to create a `template` package for them
- We will continue to use the following syntax and casing for variables in template files: `{my-variable-name}`

#### Encapsulation

- Files should encapsulate components of business logic as much as possible
  - We had a situation where `DeployNPM` used `Deployer.CommandLineParams` to parse CLI arguments, before actually calling `Deployer.deploy()` to run the core code. This indicated that `CommandLineParams` should live closer to `DeployNPM` and not be scoped to `Deployer`. In the end, it was moved to `Options`, which makes sense as options are naturally sourced from CLI parameter values.
- Even within one package, files' dependencies should ideally form an acyclic graph
  - We had a situation where `DeployNPM` and `Deployer` both depended on each other, which made things harder to reason about. To resolve the cycle, we introduced the `Options` file.

#### Robustness

- We should not assume a method from a 3rd-party library is idempotent - if in doubt, store the result to a variable and reuse it

PR #334, opened 2022-01-18
Mask auth token parameter when running deploy_npm

alexjpwalker:

The auth token passed to the `deploy_npm` rule is now masked in the rule's output logs.

Mask auth token parameter when running deploy_npm


npm/deploy/Deployer.kt
```diff
@@ -26,6 +26,7 @@ import com.vaticle.bazel.distribution.common.Logging.LogLevel.DEBUG
 import com.vaticle.bazel.distribution.common.OS.MAC
```
alexjpwalker, line 26: Update `deploy_npm` implementation, marking the auth token as a non-printable parameter, so it will be masked in the rule output.

lolski (approved):

PR #336, opened 2022-02-18
Better support for multiplatform Maven publications w/ local JARs

### LINKED ISSUE #335: Support for bundling local JARs
In Gradle, for local vendor JARs, we would zip them in the JAR to publish, such that the published JAR had what it needed to compile and be included as a dependency for other modules:

```kotlin
tasks.named<Jar>("jar") {
    from(localLibs.map {
        if (it.isDirectory) it else zipTree(it)
    })
}
```

Trying to set that up with Bazel looks like this:

```bazel
java_import(
    name = "local_jar",
    jars = [
        ":local-jar.jar",
    ],
)

java_library(
    name = "my_lib",
    srcs = glob(["src/main/**/*"])
    exports = [
        ":local_jar",
    ],
)

assemble_maven(
    name = "my_lib_assemble",
    target = ":my_lib",
    version_file = "//:VERSION",
    workspace_refs = "@plugin_workspace_refs//:refs.json",
)

deploy_maven(
    name = "my_lib_deploy",
    target = "my_lib_assemble",
    release = "...",
    snapshot = "...",
)
```

I get an error about when the `deploy_maven` rule tries to aggregate dependency info:

```
ERROR: BUILD: in @vaticle_bazel_distribution//maven:rules.bzl%aggregate_dependency_info aspect on java_import rule //:local_jar: 
	File "/external/vaticle_bazel_distribution/maven/rules.bzl", line 199, column 78, in _aggregate_dependency_info_impl
		class_jar = target[OutputGroupInfo].compilation_outputs.to_list()[0],
Error: index out of range (index is 0, but sequence has 0 elements)
```

sugarmanz:

Support bundling local `java_import` deps that may or may not conflict when bundled (#335).

The use case here is around bundling JARs for different platforms and potentially having a multiplatform JAR as well. 

1. Support for bundling `java_import` targets

Add a bit more graceful handling for dependencies that don't have a `maven_coordinates` tag. Originally, the aggregation would attempt to just pull directly from `target[OutputGroupInfo].compilation_outputs` without checking to see if it actually existed, resulting in the error from #335. Now, that logic is guarded in checking that `compilation_outputs` is truthy. Additionally, the source JAR is optional in `java_import`, so I added some safety against adding the source JAR when it doesn't exist.

Just that alone didn't fix `java_import` targets though, as `java_import` doesn't treat the input JAR as `compilation_outputs`. Locally, I had wrapped the `java_import` rule to copy over `target[JavaInfo].runtime_output_jars` into `target[OutputGroupInfo].compilation_outputs`, but requiring consumers to have to do that isn't great. So, I added another branch to the aggregation method, that checks for `target[JavaInfo].runtime_output_jars` and will create a collection of dependencies if truthy. Because `java_import` can accept multiple JARs, we need to ensure that a dependency struct for each will be created, potentially mapping to a source jar, if also defined for the index.

There are technically two different fields from `JavaInfo` that would've worked for `java_import`:
[runtime_output_jars](https://bazel.build/rules/lib/JavaInfo?hl=en#runtime_output_jars)
[full_compile_jars](https://bazel.build/rules/lib/JavaInfo?hl=en#full_compile_jars)

I used `runtime_output_jars` only because that made a _little_ more sense to me? To bundle the runtime JAR vs the compile JAR. I'm not really sure, but there is no difference between the two for `java_import`s. Open to input here.

<details>
<summary>Provider info for a `java_import` target</summary>

```
    "JavaInfo": struct(
        annotation_processing = None,
        api_generating_plugins = <unknown object com.google.devtools.build.lib.rules.java.AutoValue_JavaPluginInfo_JavaPluginData>,
        compilation_info = <unknown object com.google.devtools.build.lib.rules.java.JavaCompilationInfoProvider>,
        compile_jars = depset([
            <generated file j2v8/libs/_ijar/j2v8_empty/j2v8/libs/j2v8_empty-6.1.0-ijar.jar>,
            <generated file j2v8/libs/_ijar/j2v8_empty/j2v8/libs/j2v8_macos_x86_64-6.1.0-ijar.jar>
        ], order = "preorder"),
        full_compile_jars = depset([
            <source file j2v8/libs/j2v8_empty-6.1.0.jar>,
            <source file j2v8/libs/j2v8_macos_x86_64-6.1.0.jar>
        ], order = "preorder"),
        java_outputs = [
            <unknown object com.google.devtools.build.lib.rules.java.AutoValue_JavaRuleOutputJarsProvider_JavaOutput>,
            <unknown object com.google.devtools.build.lib.rules.java.AutoValue_JavaRuleOutputJarsProvider_JavaOutput>
        ],
        outputs = <unknown object com.google.devtools.build.lib.rules.java.JavaRuleOutputJarsProvider>,
        plugins = <unknown object com.google.devtools.build.lib.rules.java.AutoValue_JavaPluginInfo_JavaPluginData>,
        runtime_output_jars = [
            <source file j2v8/libs/j2v8_empty-6.1.0.jar>,
            <source file j2v8/libs/j2v8_macos_x86_64-6.1.0.jar>
        ],
        source_jars = [<source file j2v8/libs/srcs.jar>],
        transitive_compile_time_jars = depset([
            <generated file j2v8/libs/_ijar/j2v8_empty/j2v8/libs/j2v8_empty-6.1.0-ijar.jar>,
            <generated file j2v8/libs/_ijar/j2v8_empty/j2v8/libs/j2v8_macos_x86_64-6.1.0-ijar.jar>
        ], order = "preorder"),
        transitive_deps = depset([
            <generated file j2v8/libs/_ijar/j2v8_empty/j2v8/libs/j2v8_empty-6.1.0-ijar.jar>,
            <generated file j2v8/libs/_ijar/j2v8_empty/j2v8/libs/j2v8_macos_x86_64-6.1.0-ijar.jar>
        ], order = "preorder"),
        transitive_native_libraries = depset([], order = "topological"),
        transitive_runtime_deps = depset([
            <source file j2v8/libs/j2v8_empty-6.1.0.jar>,
            <source file j2v8/libs/j2v8_macos_x86_64-6.1.0.jar>
        ], order = "preorder"),
        transitive_runtime_jars = depset([
            <source file j2v8/libs/j2v8_empty-6.1.0.jar>,
            <source file j2v8/libs/j2v8_macos_x86_64-6.1.0.jar>
        ], order = "preorder"),
        transitive_source_jars = depset([<source file j2v8/libs/srcs.jar>])
    ),
    "OutputGroupInfo": struct(
        _direct_source_jars = depset([<source file j2v8/libs/srcs.jar>]),
        _hidden_top_level_INTERNAL_ = depset([<source file j2v8/libs/j2v8_empty-6.1.0.jar>, <source file j2v8/libs/j2v8_macos_x86_64-6.1.0.jar>]),
        _source_jars = depset([<source file j2v8/libs/srcs.jar>]),
        compilation_outputs = depset([])
    )
```
</details>

Finally, the last thing required for local JARs was ensuring that multiple targets aren't bundling the local JAR multiple times. The strategy here was to filter out transitive JAR dependencies from dependencies that have Maven coordinates, since the local JAR would have already been bundled within that dependency and can be pulled in from a Maven repository.

https://github.com/sugarmanz/bazel-distribution/blob/412162f901f6650e67e014156945ee06c6619b1b/maven/rules.bzl#L223-L229

2. ~~Add flag for allowing conflicts during JAR assembly~~

~~In the use case above, I mentioned publishing platform specific JARs, of which each bundle their own copy of a platform specific local JAR that includes the classes and platform binary for that lib. I also need to publish an "uber" JAR that contains all the binaries for each platform. When attempting to assemble the JAR for this, the JAR assembler would throw an error for duplicate entries. This makes sense for the primary use case, but in this case, I know that the classes are the same and would like to ignore the duplicates, as is currently done when I build my project with Gradle. Thus, I added a flag to not throw when duplicates are detected. This defaults to `True`, ensuring the same behavior unless specifically overridden by the consumer.~~

Shelved because this issue was inherently fixed with:
https://github.com/sugarmanz/bazel-distribution/blob/412162f901f6650e67e014156945ee06c6619b1b/maven/rules.bzl#L223-L229

3. Scope generated `deploy.py`

This essentially enables the `deploy_maven` rule to be used multiple times in the same build file, which is desirable for my mulitplatform project, as I want to have all my deploy targets for that module in the same build file.

Closes #335

### DIFF
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -58,6 +58,9 @@ class JarAssembler : Callable<Unit> {
     @Option(names = ["--jars"], split = ";")
     lateinit var jars: Array<File>
 
+    @Option(names = ["--fail_on_duplicate_entry"])
+    lateinit var failOnDuplicateEntry: Boolean
+
     private val entries = HashMap<String, ByteArray>()
     private val entryNames = mutableSetOf<String>()
 
@@ -72,7 +75,9 @@ class JarAssembler : Callable<Unit> {
                 ZipFile(jar).use { jarZip ->
                     jarZip.entries().asSequence().forEach { entry ->
                         if (entryNames.contains(entry.name)) {
-                            throw RuntimeException("duplicate entry in the JAR: ${entry.name}")
+                            if (failOnDuplicateEntry)
+                                throw RuntimeException("duplicate entry in the JAR: ${entry.name}")
+                            return@forEach
                         }
                         if (entry.name.contains("META-INF")) {
                             // pom.xml will be added by us
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -59,7 +59,7 @@ class JarAssembler : Callable<Unit> {
     lateinit var jars: Array<File>
 
     @Option(names = ["--fail_on_duplicate_entry"])
-    lateinit var failOnDuplicateEntry: Boolean
+    var failOnDuplicateEntry: Boolean = True
 
     private val entries = HashMap<String, ByteArray>()
     private val entryNames = mutableSetOf<String>()
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -58,9 +58,6 @@ class JarAssembler : Callable<Unit> {
     @Option(names = ["--jars"], split = ";")
     lateinit var jars: Array<File>
 
-    @Option(names = ["--fail_on_duplicate_entry"])
-    var failOnDuplicateEntry: Boolean = true
-
     private val entries = HashMap<String, ByteArray>()
     private val entryNames = mutableSetOf<String>()
 
@@ -75,9 +72,7 @@ class JarAssembler : Callable<Unit> {
                 ZipFile(jar).use { jarZip ->
                     jarZip.entries().asSequence().forEach { entry ->
                         if (entryNames.contains(entry.name)) {
-                            if (failOnDuplicateEntry)
-                                throw RuntimeException("duplicate entry in the JAR: ${entry.name}")
-                            return@forEach
+                            throw RuntimeException("duplicate entry in the JAR: ${entry.name}")
                         }
                         if (entry.name.contains("META-INF")) {
                             // pom.xml will be added by us
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -59,7 +59,7 @@ class JarAssembler : Callable<Unit> {
     lateinit var jars: Array<File>
 
     @Option(names = ["--fail_on_duplicate_entry"])
-    var failOnDuplicateEntry: Boolean = True
+    var failOnDuplicateEntry: Boolean = true
 
     private val entries = HashMap<String, ByteArray>()
     private val entryNames = mutableSetOf<String>()
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -186,7 +186,7 @@ def _aggregate_dependency_info_impl(target, ctx):
     deps = getattr(ctx.rule.attr, "deps", [])
     runtime_deps = getattr(ctx.rule.attr, "runtime_deps", [])
     exports = getattr(ctx.rule.attr, "exports", [])
-    deps_all = deps + exports + runtime_deps
+    deps_all = map(lambda target: target[JarInfo], deps + exports + runtime_deps)
 
     maven_coordinates = find_maven_coordinates(target, tags)
     dependencies = []
@@ -218,24 +218,15 @@ def _aggregate_dependency_info_impl(target, ctx):
             jars, source_jars + [None] * (len(jars) - len(source_jars))
         )]
 
-    print("1")
-    print([target for target in deps_all])
-    print("2")
-    print([target[JarInfo] for target in deps_all])
-    print("3")
-    print([target[JarInfo].deps for target in deps_all])
-    print("4")
-    print([target for target in deps_all if target[JarInfo].name])
-    print("5")
-    print([target[JarInfo] for target in deps_all if target[JarInfo].name])
-    print("6")
-    print([target[JarInfo].deps for target in deps_all if target[JarInfo].name])
-    print("end")
-
     return JarInfo(
         name = maven_coordinates,
-        # Not entirely sure why this doesn't work.... trying to exclude transitive deps that don't have a maven coordinate
-        deps = depset(dependencies, transitive = [target[JarInfo].deps for target in deps_all if target[JarInfo].name]),
+        deps = depset(dependencies, transitive = [
+            # Filter transitive JARs from dependency that has maven coordinates
+            # because those dependencies will already include the JARs as part
+            # of their classpath
+            depset([dep for dep in target.deps.to_list() if dep.type == 'pom'])
+                if target.name else target.deps for target in deps_all
+        ]),
     )
 
 aggregate_dependency_info = aspect(
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -205,7 +205,7 @@ def _aggregate_dependency_info_impl(target, ctx):
 
     return JarInfo(
         name = maven_coordinates,
-        deps = depset([dependency], transitive = [target[JarInfo].deps for target in deps_all]),
+        deps = depset([dependency] if dependency else [], transitive = [target[JarInfo].deps for target in deps_all]),
     )
 
 aggregate_dependency_info = aspect(
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -196,11 +196,12 @@ def _aggregate_dependency_info_impl(target, ctx):
             maven_coordinates = maven_coordinates
         )
     elif target[OutputGroupInfo].compilation_outputs:
+        source_jars = target[OutputGroupInfo]._source_jars.to_list()
         # include in the JAR
         dependency = struct(
             type = "jar",
             class_jar = target[OutputGroupInfo].compilation_outputs.to_list()[0],
-            source_jar = target[OutputGroupInfo]._source_jars.to_list()[-1],
+            source_jar = source_jars[-1] if source_jars else None,
         )
 
     return JarInfo(
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -210,12 +210,11 @@ def _aggregate_dependency_info_impl(target, ctx):
     elif target[JavaInfo].runtime_output_jars:
         jars = target[JavaInfo].runtime_output_jars
         source_jars = target[JavaInfo].source_jars
-        create_dependency = lambda jar, source_jar: struct(
+        dependencies = [struct(
             type = "jar",
             class_jar = jar,
             source_jar = source_jar,
-        )
-        dependencies = [create_dependency(dep) for dep in zip(
+        ) for (jar, source_jar) in zip(
             jars, source_jars + [None] * (len(jars) - len(source_jars))
         )]
 
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -218,13 +218,9 @@ def _aggregate_dependency_info_impl(target, ctx):
             jars, source_jars + [None] * (len(jars) - len(source_jars))
         )]
 
-    print([target[JarInfo] for target in deps_all])
-    print("hey")
-    print([target[JarInfo].deps for target in deps_all])
-
     return JarInfo(
         name = maven_coordinates,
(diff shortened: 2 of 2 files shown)


alexjpwalker:
Thanks for the contribution!

Regarding point no.2, we've never (legitimately) encountered the problem with duplicate JAR entries, presumably because we haven't tried compiling a multiplatform JAR yet. Could you provide an example / code snippet / log output to show how the problem occurs?

sugarmanz:
> Regarding point no.2, we've never (legitimately) encountered the problem with duplicate JAR entries, presumably because we haven't tried compiling a multiplatform JAR yet. Could you provide an example / code snippet / log output to show how the problem occurs?

Some news on this front, I'm not entirely sure this is required anymore. I had been hitting this case:
https://github.com/sugarmanz/bazel-distribution/blob/412162f901f6650e67e014156945ee06c6619b1b/maven/JarAssembler.kt#L79

Because of this comment:
https://github.com/vaticle/bazel-distribution/pull/336#issuecomment-1045402104

When I solved for that, it removed the JAR from being included multiple times because the intermediate platform JARs were published separately and classpath duplicates could then just be resolved at compile-time.

That being said, I'm not necessarily finding it hard to envision a scenario where this might be a problem again, like if I wasn't going to publish the intermediate JARs, such that they couldn't be included via the POM and the `JarAssembler` would be attempting to copy the same files from different intermediate JARs. Maybe that's a shortcoming of my approach for filtering out local JARs from being bundled. But as it stands, this branch works (w/o disabling errors on conflicts), so if we want to shelve that config for another PR or wait until the use case comes up again, we can.

maven/JarAssembler.kt
```diff
@@ -58,6 +58,9 @@ class JarAssembler : Callable<Unit> {
     @Option(names = ["--jars"], split = ";")
     lateinit var jars: Array<File>
 
+    @Option(names = ["--fail_on_duplicate_entry"])
+    var failOnDuplicateEntry: Boolean = true
```
alexjpwalker, line 62: Given that the duplicate JAR entry problem has now been solved by ensuring the duplicates don't make it in in the first place, I think we should err on the side of caution and remove this `failOnDuplicateEntry` option for now.

maven/rules.bzl
```diff
@@ -187,25 +189,44 @@ def _aggregate_dependency_info_impl(target, ctx):
...
-        dependency = struct(
+        dependencies = [struct(
             type = "jar",
             class_jar = target[OutputGroupInfo].compilation_outputs.to_list()[0],
-            source_jar = target[OutputGroupInfo]._source_jars.to_list()[-1],
-        )
+            source_jar = source_jars[-1] if source_jars else None,
+        )]
+    # include runtime output jars
+    elif target[JavaInfo].runtime_output_jars:
```
alexjpwalker, line 210: I really like this solution - it makes perfect sense that `compilation_outputs` would be undefined for a `java_import` target, given that the JAR(s) are already compiled!

Can we maybe remove the `elif target[OutputGroupInfo].compilation_outputs` branch entirely? Are there any cases when `runtime_output_jars` would be undefined? I note you proposed `full_compile_jars` as an alternative, so that is an option to consider as well.

maven/rules.bzl
```diff
@@ -187,25 +189,44 @@ def _aggregate_dependency_info_impl(target, ctx):
...
     return JarInfo(
         name = maven_coordinates,
-        deps = depset([dependency], transitive = [target[JarInfo].deps for target in deps_all]),
+        deps = depset(dependencies, transitive = [
+            # Filter transitive JARs from dependency that has maven coordinates
+            # because those dependencies will already include the JARs as part
+            # of their classpath
+            depset([dep for dep in target[JarInfo].deps.to_list() if dep.type == 'pom'])
+                if target[JarInfo].name else target[JarInfo].deps for target in deps_all
+        ]),
```
alexjpwalker, line 229: Can you create a reproducible example to demonstrate this problem? All targets in https://github.com/sugarmanz/bazel-distribution-local-jars appear to build OK without this particular change.

maven/rules.bzl
```diff
@@ -297,7 +322,7 @@ MavenDeploymentInfo = provider(
 
 
 def _deploy_maven_impl(ctx):
-    deploy_maven_script = ctx.actions.declare_file("deploy.py")
+    deploy_maven_script = ctx.actions.declare_file("%s-deploy.py" % ctx.attr.name)
```
alexjpwalker, line 325: This change looks good to me.

alexjpwalker (changes requested):

maven/JarAssembler.kt
sugarmanz: Agreed. Will remove from this PR and will re-open if I end up needing this again, but from what it seems like, the use case stems from an edge-case project structure and could be avoided by restructuring the project.

maven/rules.bzl
sugarmanz: > All targets in https://github.com/sugarmanz/bazel-distribution-local-jars appear to build OK without this particular change.

In the sense that Bazel is able to build them successfully, yes. But the desired output is not met. Before this change, `lib-b` would bundle `lib-v` even though it is a transitive dependency that is already bundled and published in `lib-a`:

`lib-a.jar`:
Screen Shot 2022-03-25 at 1 00 02 PM

`lib-b.jar`:
Screen Shot 2022-03-25 at 12 53 33 PM

With this change, it filters out things that will get bundled into the JAR from the transitive dependencies, resulting in `lib-b` no including `lib-v` files as part of the classpath.

`lib-a.jar`:
Screen Shot 2022-03-25 at 12 53 12 PM

`lib-b.jar`:
Screen Shot 2022-03-25 at 12 59 49 PM

Here are the JARs built against commit `bf78b8a` (before that change) of my branch:

https://github.com/sugarmanz/bazel-distribution-local-jars/tree/main/jars/2

Here are the JARs built against latest (`f0be8da`):

https://github.com/sugarmanz/bazel-distribution-local-jars/tree/main/jars/desired

maven/rules.bzl
sugarmanz: > Can we maybe remove the elif target[OutputGroupInfo].compilation_outputs branch entirely?

That's a really good question. The main reason I didn't remove it was because `target[OutputGroupInfo].compilation_outputs` seemed to fit the more generic case, given that it uses the standard `OutputGroupInfo` provider. Adding another branch to catch `JavaInfo` specific providers originally was intended to narrow down and catch the edge cases for rules that don't expose `compilation_outputs` as part of  `target[OutputGroupInfo]`.

> it makes sense that compilation_outputs would be undefined for a java_import target, given that the JAR(s) are already compiled!

That being said, I think maybe the using the `JavaInfo` provider might just be the more targeted/correct version of the `OutputGroupInfo` branch, and given that this is a rule for publishing Java artifacts to some Maven repository, it might make sense _just_ to have that branch. It is a bit more sophisticated in how it gathers the compile and source JARs, which would support some cases not previously supported.

I did a test to verify if we could remove the "generic" branch, and if that would even support bundling `java_import` and other `kt_jvm_library` targets and it does. My main apprehension if someone is depending on bundling targets that don't provide `JavaInfo`. But in that case, they are the edge case, so at the very least, we could swap the branches such that the more sophisticated handling is done before the generic catch-all.

Thoughts?

alexjpwalker: I'm also thinking we take a bold step and remove the branch with `compilation_outputs` on the grounds that
1. After some internal testing, we can't find anywhere where it's necessary
2. In its use case, it is probably a suboptimal approach

So if it does cause a regression for anyone, we should be able to write a fix for it that is better than the current code.

maven/rules.bzl
alexjpwalker: Understood. This is a good change.

maven/rules.bzl
sugarmanz: Awesome. Makes sense to me. I'll remove that branch and push it up!

alexjpwalker: Just one last thing:
```suggestion
    else:
```
If there is a bug in our implementation, this change will ensure we throw an error rather than just keeping silent about it.

sugarmanz: Yeah, makes sense. I added an explicit `fail` branch to ensure that it's clear why it failed:

https://github.com/sugarmanz/bazel-distribution/blob/a694a8c2c1ebb95cd2e52f37843b1aed2eeba915/maven/rules.bzl#L209-L210

alexjpwalker: That's even better 👍

alexjpwalker:
PR merged - thank you for the contribution!

lolski:
Thank you @sugarmanz ! Great work there

PR #337, opened 2022-02-23
Make assemble_npm support M1 architecture

adierkens:

To fix issues with running the assembly rules with a m1 mac

- Adds the correct architecture folder under the `PATH` search


alexjpwalker:
Thanks for the contribution!

Is this change (adding `external/nodejs_darwin_arm64/bin/` to the `PATH`) sufficient to support M1 architecture or is there more to come?

adierkens:
There was 1 other issue I was hitting earlier, but I believe that's unrelated to this repo. This should be sufficient

alexjpwalker (approved):
Looks good to me!

PR #339, opened 2022-03-29
DRAFT: java_deps: Fail if we find duplicate Maven coordinates

alexjpwalker:

{ Please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above, and reference the GitHub issues to be automatically closed, such like 'closes #number'. }

### DIFF
--- a/common/java_deps/rules.bzl
+++ b/common/java_deps/rules.bzl
@@ -113,6 +113,13 @@ def _java_deps_impl(ctx):
     outputPathOverrides = ctx.attr.java_deps_root_overrides
 
     mapping = ctx.attr.target[TransitiveJarToMavenCoordinatesMapping].mapping
+    distinct_mapping = {}
+
+    for k, v in mapping.items():
+        if v in distinct_mapping.values() and k not in distinct_mapping.keys():
+            conflicting_file_name = [a for (a, b) in distinct_mapping.items() if b == v][0]
+            fail("{} and {} both have the same Maven coordinates, {}. Distinct JARs must have distinct Maven coordinates.".format(conflicting_file_name, k, v))
+        distinct_mapping[k] = v
 
     for file in ctx.attr.target.data_runfiles.files.to_list() + ctx.attr.target.files.to_list():
         if file.extension == 'jar' and not file.path.startswith(LOCAL_JDK_PREFIX):


common/java_deps/rules.bzl
```diff
@@ -108,28 +108,30 @@ _transitive_collect_maven_coordinate = aspect(
...
     names = {}
-    files = []
-    filenames = []
-    outputPathOverrides = ctx.attr.java_deps_root_overrides
+    files_by_output_path = {}
+    output_path_overrides = ctx.attr.java_deps_root_overrides
 
     mapping = ctx.attr.target[TransitiveJarToMavenCoordinatesMapping].mapping
 
-    for file in ctx.attr.target.data_runfiles.files.to_list() + ctx.attr.target.files.to_list():
```
alexjpwalker, line 117: It looks like we added `ctx.attr.target.files` here a couple of years ago, in order to support a particular scenario that we ended up not using - so now we're cleaning it up.

common/java_deps/rules.bzl
```diff
@@ -108,28 +108,30 @@ _transitive_collect_maven_coordinate = aspect(
...
-        if file.extension == 'jar' and not file.path.startswith(LOCAL_JDK_PREFIX):
+    for file in ctx.attr.target.data_runfiles.files.to_list():
```
alexjpwalker, line 116: High level explanation:

For each JAR file, not part of the JDK, that `ctx.attr.target` imports as a runtime dependency:
1. Check that it has a Maven coordinates tag, if this `java_deps` mandates them
2. Determine the output path of this JAR. If it has Maven coordinates, replace `.` and `:` in them to `-`.
3. Add the output path and JAR to a map (of output paths to JARs). If the output filename was already taken by another JAR, throw. The values of this map are the primary outputs of this `java_deps`.
4. Determine the _full_ output path of this JAR, normally `java_deps_root + output_path + ".jar"`, but may be overridden using the `output_path_overrides` map (e.g: to enable some JARs to be included on the classpath only when certain criteria are met). This is used to output a "JAR mapping file" which can be used as a reference.

haikalpribadi (approved):

Issue #233, opened 2020-06-04
feature-request: add android_library support to maven distribution


jagobagascon:
I'm trying to deploy an `android_library` to maven, but maven assemble and deployment rules only work with jars, and I need to upload an `aar` artifact.

I tried modifying the rules myself to make a PR ~~but for some reason aspects are not working when I define the `aar` as target (and thus pom generation fails). I think [it may be a bug](https://github.com/bazelbuild/bazel/issues/11545) with Bazel or the Android rules~~ but I couldn't make it work without having to completely change the deployment rule. 

The `aar` artifact is an implicit output of the `android_library` rule, which means it's just a file and aspects do not work on them.

```python
assemble_maven(
    name = 'assemble',
    target = ':my_android_lib',
)
assemble_maven(
    name = 'assemble',
    target = ':my_android_lib.aar', # fails because aspects are not executed
)
```

To make it work I had to modify the deployment so I could generate the pom and then explicitly choose the deployed file:
```python
assemble_maven(
    name = 'generate_pom',
    target = ':build',
)
publish_maven(
    name = 'publish',
    pom_file = ':generate_pom',
    target = ':build.aar',
    sources = ':build.srcjar',
)
```

Are there any plans on supporting Android `aar` files?

lolski:
Hi! Unfortunately, we have no plans for that in the foreseeable future (we're not building Android products as a company).

What you can do instead is to create your own distribution repository, and it can depend on this repository if you feel that you can build it on top of existing functionalities which we provide.

Closing this issue for now.

cpsauer:
@jagobagascon, did you end up solving this another way? Was looking into doing the same :)

jagobagascon:
@cpsauer I ended up writing my own Bazel rule. As I say in my first comment, I had to split the deployment in two: pom generation and the deployment itself.

The rule to create the `pom.xml` does just the same as this library (using the Android library we can get the dependencies to generate the pom) and the deployment rule has a `target` field to define the deployed file. It just calls `mvn deploy:deploy-file` with the generated pom and the aar file I want to deploy.

```python
# generate pom
generate_pom(
    name = 'generate_pom',
    target = ':my_android_lib', # android_library (only used to generate the pom)
    project_url = project_url,
    project_name = 'my_android_lib',
    packaging = 'aar',
)

# publish calling `bazel run :publish --define version=1.2.3`
publish_to_artifactory(
    name = 'publish',
    pom_file = ':generate_pom',
[...]

PR #344, opened 2022-06-01
Android library publishing support

sugarmanz:

Add publishing support for `android_library` targets (#233)

First, the rules, assemblers, and scripts need to take into account other packaging types. Determining the packaging type and passing that info along via the providers was the necessary first step.

Second, the `JarAssembler.kt` file only supported packaging a `JAR`. Rather than duplicate the script for `AAR` support, I added support directly in that file to reuse a lot of the same logic used to assemble `JAR`s. There is one TODO left surrounding duplicate entries, but I believe that might be something related to my specific project. 

Third, handling the Android SDK and neverlink dependencies. In order to be able to perform the `aggregate_dependency_info` aspect, the Android SDK target needed to be handled. It is a specific rule type baked into the Bazel codebase. I wasn't really sure how to determine the rule type within the aspect, so I just hardcoded the label as something to ignore (gross, I know). Additionally, a third party ruleset I use for databinding support had an additional `java_library` that captured the Android SDK. Luckily, this was marked as `neverlink`, so I was able to key off of that attribute to avoid capturing the Android SDK from that target in the transitive closure. That should be a decent strategy to follow since the artifacts we're assembling probably shouldn't include `neverlink` dependencies that were just used to compile.

### DIFF
--- a/maven/JarAssembler.kt
+++ b/maven/JarAssembler.kt
@@ -28,6 +28,7 @@ import java.io.BufferedInputStream
 import java.io.BufferedOutputStream
 import java.io.File
 import java.io.FileOutputStream
+import java.io.InputStream
 import java.nio.charset.Charset
 import java.nio.file.Path
 import java.nio.file.Paths
@@ -37,9 +38,10 @@ import java.util.zip.ZipFile
 import java.util.zip.ZipInputStream
 import java.util.zip.ZipOutputStream
 import kotlin.RuntimeException
-import kotlin.collections.HashMap
 import kotlin.system.exitProcess
 
+typealias Entries = MutableMap<String, ByteArray>
+private fun Entries(): Entries = mutableMapOf()
 
 @Command(name = "jar-assembler", mixinStandardHelpOptions = true)
 class JarAssembler : Callable<Unit> {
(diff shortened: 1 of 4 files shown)


maven/rules.bzl
```diff
@@ -88,12 +97,14 @@ def _generate_class_jar(ctx, pom_file):
     maven_coordinates = _parse_maven_coordinates(target[JarInfo].name)
 
     jar = None
-    if hasattr(target, "files") and target.files.to_list() and target.files.to_list()[0].extension == "jar":
+    if (_is_android_library(target)):
+        jar = target[AndroidLibraryAarInfo].aar
```
alexjpwalker, line 101: The information we need here is only the path to the JAR/AAR file.

Is there any way we could retrieve this in a way that is generic to both Java JARs and Android AARs, without requiring a check for `_is_android_library`? Since ideally we would like to keep this rule as generalised as possible.

maven/rules.bzl
```diff
@@ -301,6 +327,7 @@ assemble_maven = rule(
 
 MavenDeploymentInfo = provider(
     fields = {
+        'packaging': 'The type of target to publish (jar, war, aar, etc.)',
```
alexjpwalker, line 330: Given that we are deploying either a JAR or an AAR file, I think it would be more intuitive if we simply read the extension of the file being passed into the `deploy_maven` rule (instead of relying on a new field in `MavenDeploymentInfo`) to determine the `packaging` used in `deploy.py`.

maven/JarAssembler.kt
```diff
@@ -58,53 +61,118 @@ class JarAssembler : Callable<Unit> {
...
+        entry.isDirectory -> {
+            // needed directories would be added by us
+            null
+        }
+        entry.name.contains("META-INF/maven") -> {
+            // pom.xml will be added by us
+            null
+        }
+        keys.contains(entry.name) -> {
+            // TODO: Investigate why I'm getting duplicates
```
alexjpwalker, line 155: How serious is this? Can we remove it?

maven/JarAssembler.kt
```diff
@@ -28,17 +28,20 @@ import java.io.BufferedInputStream
 import java.io.BufferedOutputStream
```
alexjpwalker, line 28: It's cool that this AAR support is added with minimal code change! However, I think the domains of Java and Android are conceptually different (even though Android uses Java code extensively.) As such, I would propose:

- Rename this file (and class) to `Assembler`. Given that we are in the `maven` package, that's unambiguous. `Assembler` should contain all the logic common to both JAR and AAR.
- Introduce two inner classes of `Assembler`: `Jar` and `Aar`, which encapsulate the JAR and AAR-specific code. (They will, presumably, need a reference to the `Assembler` class in their constructors.)

alexjpwalker (changes requested):

sugarmanz:
Apologies for the delay. I'll be able to revisit this PR in a little while. Just want to reassure that it is my goal to get this merged to help provide broader publishing support to the community 😄

PR #346, opened 2022-07-15
[assemble-pip] Add python_requires to limit package to python 3 only

caseyduquettesc:

Produce pip distributions that aren't available to python 2 users (or any arbitrary python version)

Adds a `python_requires` attribute to `assemble_pip` passed through to `setup.py` as defined in https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

For example,
```py
assemble_pip(
    ...
    python_requires = ">=3",
)
```
should produce a setup.py that makes the package unavailable to users of python 2.

Also fixes an error in the assemble script if the package is empty.

### DIFF
--- a/pip/assemble.py
+++ b/pip/assemble.py
@@ -57,22 +57,23 @@ args.imports = list(map(
 # new package root
 pkg_dir = tempfile.mkdtemp()
 
+if not args.files:
+    raise Exception("Cannot create an archive without any files")
 
-if args.files:
-    for f in args.files:
-        fn = f
-        for _imp in args.imports:
-            match = _imp.match(fn)
-            if match:
-                fn = match.group('fn')
-                break
-        try:
-            e = os.path.join(pkg_dir, os.path.dirname(fn))
-            os.makedirs(e)
-        except OSError:
-            # directory already exists
-            pass
-        shutil.copy(f, os.path.join(pkg_dir, fn))
+for f in args.files:
+    fn = f
+    for _imp in args.imports:
+        match = _imp.match(fn)
+        if match:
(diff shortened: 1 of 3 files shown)


pip/assemble.py
```diff
@@ -58,20 +58,21 @@ def create_init_files(directory):
...
-            fn = match.group('fn')
-            break
-    try:
-        e = os.path.join(pkg_dir, os.path.dirname(fn))
-        os.makedirs(e)
-    except OSError:
-        # directory already exists
-        pass
-    shutil.copy(f, os.path.join(pkg_dir, fn))
+if args.files:
```
alexjpwalker, line 61: Is there ever a scenario in which `args.files` would be legitimately empty? Does it make sense to assemble a Pip package with no Python sources?

If so, we should keep this change; if not, it would be better to raise an error when `args.files` is None or empty.

alexjpwalker:
Thanks for the contribution, @caseyduquettesc !

I've left one comment on the code, the change itself looks like a good feature to support.

pip/assemble.py
caseyduquettesc: I think it's technically allowed in the wheel spec, it's likely undesirable from a user's perspective. The error that was thrown didn't make it clear the cause was missing all input files. That might be the better route to take here actually. Something like
```py
if not args.files:
    raise Exception("Cannot create an archive without any files")
```

alexjpwalker: Agreed - let's add that.

alexjpwalker (approved):
Looks good to me.

PR #347, opened 2022-07-15
[assemble-pip] Support stripping a path prefix from package files

caseyduquettesc:

In a monorepo with many python packages, where each is individually published, the libraries might be a directory or two lower in the repo than what you want to be published.

For example, take this folder structure

```
javacode/
pybindings/
    apiclient/
        BUILD.bazel
        __init__.py
````

All the python packages are under `pybindings`, but I don't want the root folder showing up in the users' imports.

Currently, the assembled file persists the `pybindings/apiclient` structure. I'd like to propose a way to strip `pybindings/` in this case and assemble a file starting at `apiclient`.

Adds `strip_prefix` to `assemble_pip`. When the input files are copied, if there's a `strip_prefix`, and a file starts with that prefix, it is stripped from the start of the path.

### DIFF
--- a/pip/assemble.py
+++ b/pip/assemble.py
@@ -43,6 +43,7 @@ parser.add_argument('--requirements_file', help="install_requires")
 parser.add_argument('--readme', help="README file")
 parser.add_argument('--files', nargs='+', help='Python files to pack into archive')
 parser.add_argument('--imports', nargs='+', help='Folders considered to be source code roots')
+parser.add_argument('--strip_prefix', help="Path prefix to strip from all package files")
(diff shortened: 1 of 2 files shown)


alexjpwalker:
Thanks for the PR @caseyduquettesc !

According to the following PR:
- https://github.com/vaticle/bazel-distribution/pull/148

`assemble_pip` had a `package_root` field back in 2019, but it turned out to be possible to reliably detect the package root by searching the target's `imports` exposed by the `PyInfo` provider.

Is this applicable in your case?

PR #349, opened 2022-07-20
[deploy_maven] GPG passphrase

sugarmanz:

Support using GPG keys that are passphrase protected

Pulls passphrase from `DEPLOY_MAVEN_GPG_PASSPHRASE` environment variable

### DIFF
--- a/maven/templates/deploy.py
+++ b/maven/templates/deploy.py
@@ -53,15 +53,20 @@ def upload(url, username, password, local_fn, remote_fn):
             local_fn, upload_status_code))
 
 
-def sign(fn):
+def sign(fn, passphrase=None):
     # TODO(vmax): current limitation of this functionality
     # is that gpg key should already be present in keyring
-    # and should not require passphrase
     asc_file = tempfile.mktemp()
     sp.check_call([
         'gpg',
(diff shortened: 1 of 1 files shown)


maven/templates/deploy.py
```diff
@@ -132,18 +139,18 @@ def unpack_args(_, a, b=False):
 
 upload(maven_url, username, password, jar_path, filename_base + '.jar')
 if should_sign:
-    upload(maven_url, username, password, sign(jar_path), filename_base + '.jar.asc')
+    upload(maven_url, username, password, sign(jar_path, gpg_passphrase), filename_base + artifact_extension + '.asc')
```
alexjpwalker, line 142: It looks like this change depends on:
- https://github.com/vaticle/bazel-distribution/pull/344

We can either merge this PR by reverting to the following:
```suggestion
    upload(maven_url, username, password, sign(jar_path, gpg_passphrase), filename_base + '.jar.asc')
```
Or wait until https://github.com/vaticle/bazel-distribution/pull/344 is merged.

alexjpwalker (changes requested):
Yes, this looks like something we should add support for.

Left one comment.

PR #350, opened 2022-07-20
[deploy_maven] allow for release repo override via command line

sugarmanz:

Allow for the release repo to be configured dynamically. With Maven Central, there is the opportunity to batch a release with a singular staging repository. Unfortunately, without some rework, it would be a little difficult to dynamically set the release repository. I don't really want to dynamically touch the `BUILD` or `.bzl` files where I might have the `release_repo` defined that is passed into the rule since that could differ between repos. This change allows the `release_repo` to be overridden from a CLI such that the staging repository can be created dynamically during CI and passed to the deploy target for publishing before dynamically closing & releasing the staging repo.

This is an interesting PR b/c there are a few ways to solve this. I could have modified the `.py` script to try and read from an env var. However, given that this is really a one-off URL, the potential for a stale repo being stored and used from the env was non-zero. Having to manually specify the staging repo w/in the bazel invocation seemed like a good middleground.

Opinions appreciated 😄

### DIFF
--- a/maven/rules.bzl
+++ b/maven/rules.bzl
@@ -17,6 +17,7 @@
 # under the License.
 #
 
+_DEPLOY_MAVEN_RELEASE_REPO_KEY = "DEPLOY_MAVEN_RELEASE_REPO"
 
 def _parse_maven_coordinates(coordinates_string, enforce_version_template=True):
     coordinates = coordinates_string.split(':')
@@ -315,6 +316,10 @@ def _deploy_maven_impl(ctx):
     src_jar_link = "lib.srcjar"
     pom_xml_link = ctx.attr.target[MavenDeploymentInfo].pom.basename
 
+    release_repo = ctx.attr.release
+    if _DEPLOY_MAVEN_RELEASE_REPO_KEY in ctx.var:
+        release_repo = ctx.var[_DEPLOY_MAVEN_RELEASE_REPO_KEY]
+
     ctx.actions.expand_template(
         template = ctx.file._deployment_script,
         output = deploy_maven_script,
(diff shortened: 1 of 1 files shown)


maven/rules.bzl
```diff
@@ -315,6 +316,10 @@ def _deploy_maven_impl(ctx):
     src_jar_link = "lib.srcjar"
     pom_xml_link = ctx.attr.target[MavenDeploymentInfo].pom.basename
 
+    release_repo = ctx.attr.release
+    if _DEPLOY_MAVEN_RELEASE_REPO_KEY in ctx.var:
+        release_repo = ctx.var[_DEPLOY_MAVEN_RELEASE_REPO_KEY]
```
alexjpwalker, line 321: I agree with the approach of using `--define`! This is aligned with how we override the `version` being deployed:
```sh
bazel run --define version=$(cat VERSION) //:deploy-maven -- release
```
A couple of points:
1. Since we're allowing users to override the release repo, we should also allow them to override the snapshot repo.
2. As `version` is all lowercase, so should be `deploy_maven_release_repo`; and for terseness, I'd call it just `release_repo`. (As this is a `run` target, the key need not be globally unique.)

With all that in mind, and simplifying the Starlark a little:
```suggestion
    snapshot_repo = ctx.var.get("snapshot_repo", default=ctx.attr.snapshot)
    release_repo = ctx.var.get("release_repo", default=ctx.attr.release) 
```
(requires additional update to L330: `"{snapshot}": snapshot_repo,`)

alexjpwalker (changes requested):

PR #367, opened 2023-01-18
Support proc macros and fix dependency name detection in `assemble_crate`

alexjpwalker:

`assemble_crate` now supports procedural macro dependencies, and is more reliable at correctly detecting the crate names of all dependencies. In particular, crates with hyphens (such as `async-trait`) should no longer cause issues.

Previously `assemble_crate` aggregated dependency info via only the `deps` attribute of a `rust_*` target. Now, `proc_macro_deps` is traversed as well.

Also, crate name detection has been reworked. Because `rules_rust` doesn't expose the original name of the crate downloaded from Cargo anywhere, we were unable to reliably get that crate name and put it into `Cargo.toml`. Now, we use string parsing against the target's label. This works because the label is (e.g):
```
@crates__async-trait-0.36.0//:async_trait
```
We take everything up to + excluding the first `.`, then take everything up to + excluding the last `-`, then strip `@crates__` from the front.

It looks like a dumb hack (and it is), but it works. I've raised:
- https://github.com/bazelbuild/rules_rust/issues/1783

### DIFF
--- a/crates/rules.bzl
+++ b/crates/rules.bzl
@@ -113,21 +113,31 @@ CrateSummary = provider(
     },
 )
 
+def _is_universe_crate(target):
+    return str(target.label).startswith("@crates__")
+
+def _universe_crate_name(target):
+    return str(target.label).split(".")[0].rsplit("-", 1)[0].removeprefix("@crates__")
+
 def _aggregate_crate_summary_impl(target, ctx):
-    name = ctx.rule.attr.name
-    for tag in ctx.rule.attr.tags:
-        if tag.startswith("crate-name"):
-            name = tag.split("=")[1]
+    if _is_universe_crate(target):
+        name = _universe_crate_name(target)
+    else:
+        name = ctx.rule.attr.name
+        for tag in ctx.rule.attr.tags:
+            if tag.startswith("crate-name"):
+                name = tag.split("=")[1]
     return CrateSummary(
         name = name,
         version = ctx.rule.attr.version,
-        deps = [target for target in getattr(ctx.rule.attr, "deps", [])]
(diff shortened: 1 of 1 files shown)


crates/rules.bzl
```diff
@@ -113,21 +113,31 @@ CrateSummary = provider(
     },
 )
 
+def _is_universe_crate(target):
+    return str(target.label).startswith("@crates__")
+
+def _universe_crate_name(target):
+    return str(target.label).split(".")[0].rsplit("-", 1)[0].removeprefix("@crates__")
```
alexjpwalker, line 120: We take everything up to + excluding the first `.`, then take everything up to + excluding the last `-`, then strip `@crates__` from the front.

crates/rules.bzl
```diff
@@ -113,21 +113,31 @@ CrateSummary = provider(
...
         version = ctx.rule.attr.version,
-        deps = [target for target in getattr(ctx.rule.attr, "deps", [])]
+        deps = [target for target in getattr(ctx.rule.attr, "deps", []) + getattr(ctx.rule.attr, "proc_macro_deps", [])]
     )
 
 
 aggregate_crate_summary = aspect(
     attr_aspects = [
        "deps",
+       "proc_macro_deps",
```
alexjpwalker, line 140: Propagate aspect via `proc_macro_deps` as well as `deps`, allowing proc-macro deps to be added to the generated `Cargo.toml`.

crates/rules.bzl
nikolaimerritt: I've taken a look at the Semantic Versioning specification. This should comply with any crate name where the name is proceeded by a valid semantic version.

nikolaimerritt (approved):

PR #368, opened 2023-01-24
Fix assemble_pip and deploy_pip

lyao-77:
Currently there are some issues with these two rules: 'assemble_pip' and 'deploy_pip'.

assemble_pip:
(1) data files are not included in "tarball" build;
(2) it cannot produce "wheel" file which is a common needs for many use cases.

deploy_pip:
(1) currently the deployment relies on "repository_registry" and username&password. However, we use "~/.pypirc" to config twine and we believe many other users have the same config. "pypi_profile" attr is added to enable this use case;
(2) this rule will run into an error due to missing deps in requirements.txt;
(3) this rule will run into an error due to that the "repo_type_key" arg in deploy.py is missing when running the script. Practically, only one of "release" and "snapshot" can be used for deployment;
(4) this rule can only deploy single package ("tar.gz") which is incompatible with what we did for "assemble_pip".

This PR is to address all of the issues above.


lyao-77:
@alexjpwalker Could you take a look at this PR?

alexjpwalker:
Thanks for the PR!

The intent looks clear and reasonable. I'll find time to review it in the coming days.

pip/requirements.txt
```diff
@@ -13,4 +13,9 @@ tqdm==4.29.1
 twine==1.12.1
 urllib3==1.26.5
 webencodings==0.5.1
-wheel==0.32.3
+wheel==0.33.1
+six==1.16.0
+charset-normalizer==2.0.12
+packaging==21.3
+pyparsing==3.0.9
```
dmitrii-ubskii, line 20: Please sort the requirements.

pip/templates/deploy.py
```diff
@@ -29,54 +29,44 @@
 # noinspection PyUnresolvedReferences
 import twine.commands.upload
 
-parser = argparse.ArgumentParser()
-parser.add_argument('repo_type')
-args = parser.parse_args()
+pypi_profile = "{pypi_profile}"
+pip_registry = "{snapshot}" if "{snapshot}" else "{release}"
+command = None
```
dmitrii-ubskii, line 34: This line is unnecessary.

pip/templates/deploy.py
```diff
@@ -29,54 +29,44 @@
...
-        pip_username,
-        '-p',
-        pip_password,
-        '--repository-url',
-        pip_registry
-    ])
+    twine.commands.upload.main(command)
 finally:
     if new_package_file:
         os.remove(new_package_file)
```
dmitrii-ubskii, line 72: We should probably remove the `new_wheel_file` here as well.

dmitrii-ubskii (changes requested):
Looks good! Please see some minor comments below while I make sure this doesn't break our deployments.

dmitrii-ubskii (changes requested):

pip/templates/deploy.py
lyao-77: I have added a new commit to address this issue.

pip/assemble.py
```diff
@@ -60,7 +63,7 @@ def create_init_files(directory):
 if not args.files:
     raise Exception("Cannot create an archive without any files")
 
-for f in args.files:
+for f in args.files + args.data_files:
```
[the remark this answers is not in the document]
lyao-77: True. I have met this error with another use case and was planning to fix.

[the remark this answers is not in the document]
lyao-77: I have added a new commit to fix this issue.

pip/templates/deploy.py
lyao-77: fixed.

pip/requirements.txt
lyao-77: fixed.

lyao-77:
> Also, it would be nice to have the option to not generate a wheel and only sdist.

Hi @dmitrii-ubskii, I have addressed all of your comments by pushing new commits. Specifically for this comment, I find it kind of difficult to make the wheel build optional because the `outputs` field of `assemble_pip` seems static and it is not supposed to be dynamic in bazel. I think it doesn't hurt to build a `.whl` file in this case.

I did push a commit to make deployment optional. So although it will build both `.whl` and `tar.gz`, the user could choose if they would like to deploy either one of them or both.

dmitrii-ubskii:
Looks good! @alexjpwalker, can you take a look?

pip/rules.bzl
```diff
@@ -288,18 +311,25 @@ deploy_pip = rule(
...
-            mandatory = True,
+            default = "",
             doc = "Remote repository to deploy pip snapshot to"
         ),
         "release": attr.string(
-            mandatory = True,
+            default = "",
             doc = "Remote repository to deploy pip release to"
         ),
+        "deploy_types": attr.string_list(
```
alexjpwalker, line 329: Hmmm.

I think, given that we _assemble_ the wheel every time, there's no additional harm in also deploying it every time. Which would reduce the number of parameters here and simplify the deployment code.

(You may have made this change - making the deployment optional - part way through developing this PR, in which case, sorry about that!)

So yeah, I think we should always deploy the wheel.

alexjpwalker (changes requested):

pip/requirements.txt
```diff
@@ -1,16 +1,31 @@
...
-idna==2.8
-pkginfo==1.5.0.1
-readme-renderer==24.0
-requests-toolbelt==0.8.0
-requests==2.27.0
-setuptools==62.1.0
-tqdm==4.29.1
-twine==1.12.1
-urllib3==1.26.5
+bleach==6.0.0
```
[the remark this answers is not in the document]
dmitrii-ubskii: Also, we notice you've updated a lot of dependencies. We try not to break compatibility without reason.

pip/BUILD
```diff
@@ -44,4 +45,4 @@ py_binary(
     name = "repackage",
     srcs = ["repackage.py"],
     visibility = ["//visibility:public"]
-)
+)
```
dmitrii-ubskii, line 48: Please ensure newlines at the end of each file!

pip/assemble.py
```diff
@@ -37,17 +37,20 @@ def create_init_files(directory):
 
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--output', help="Output archive")
+parser.add_argument('--output', '--output_sdist', help="Output sdist archive")
```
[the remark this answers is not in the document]
lyao-77: reverted the related previous commit.

pip/rules.bzl
lyao-77: reverted.

pip/BUILD
lyao-77: addressed.

pip/requirements.txt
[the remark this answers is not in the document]
lyao-77: Looks like the repo didn't implement an interpreter solution for those rules. So although there is a requirements.txt , those deps are actually not used correctly. Based on the implementation the rules will use the system  dependency rather than what it installed in the sandbox through requirements.txt . I changed a bit in `deploy.py`  to make it use the runfile dependency first to avoid the issue.  The change is included in `fix dependency issue` commit.

The reason there are many deps missing in the original requirements.txt is because the rule is using your system packages.

lyao-77:
@alexjpwalker @dmitrii-ubskii  Hi I have addressed all of the new comments. Can you take a look again? Thanks!

pip/requirements.txt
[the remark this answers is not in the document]
alexjpwalker: Good heavens, I didn't notice that little change. Thanks!

dmitrii-ubskii (approved):

flyingsilverfin:
@lyao-77 Hi, please note that we've made some changes on top of your changes to bring the rule back in line with the standard architecture we've chosen for this repository: https://github.com/vaticle/bazel-distribution/pull/373

You'll still be able to use exactly the functionality you require as implemented in this PR, just with a different API.

flyingsilverfin:
I'm actually also going to remove the `universal` part of the python wheel distribution - if this is something you would like, you should add it as an argument to `assemble_pip`. Most Python libraries these days are not Python 2 and 3 compatible, which is one of the requirements for Universal = 1: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels

flyingsilverfin:
I'm actually also going to remove the `universal` part of the python wheel distribution - if this is something you would like, you should add it as an argument to `assemble_pip`. Most Python libraries these days are not Python 2 and 3 compatible, which is one of the requirements for Universal = 1: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels

This also means we're going to fix the assemble and deploy rules to be only compatible with Python 3.x, which is implemented in #375

PR #371, opened 2023-03-02
add entry_point and setuptools-git

lyao-77:
Please replace every line in curly brackets ( { like this } ) with an appropriate description, and remove this line.

{ Please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above, and reference the GitHub issues to be automatically closed, such like 'closes #number'. }

### DIFF
--- a/pip/BUILD
+++ b/pip/BUILD
@@ -36,6 +36,7 @@ py_binary(
     srcs = ["assemble.py"],
     deps = [
         vaticle_bazel_distribution_requirement("setuptools"),
+        vaticle_bazel_distribution_requirement("setuptools-git"),
         vaticle_bazel_distribution_requirement("wheel")
     ],
     visibility = ["//visibility:public"]
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -35,3 +35,4 @@ urllib3==1.26.14
 webencodings==0.5.1
 wheel==0.38.4
 zipp==3.11.0
+setuptools-git==1.2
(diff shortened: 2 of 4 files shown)


lyao-77:
Hi @dmitrii-ubskii @alexjpwalker  Can you take a look at this PR? Thanks!

lyao-77:
Hi @flyingsilverfin , I think the `entry_point` field is only a filed for `setuptools`. We would like to add it because when we assemble our package using `setuptools` directly, we use this field.

Regarding the other change, the use case is that we used this rule to assemble one of our internal python library, and then use it in another project. We met some issues during `import` and realized that when bazel load it's dependency, it will generate a __init__.py  for each module with the following content: 

"# __path__ manipulation added by bazelbuild/rules_python to support namespace pkgs.
__path__ = __import__('pkgutil').extend_path(__path__, __name__)
"

However, since we hardcoded an empty `__init__.py` in current implementation during packaging, it will interrupt how bazel works and the `import` won't work

alexjpwalker:
Hi, thanks for the contribution! By way of update (and managing expectations), unfortunately we don't currently have the resources to test this code in detail and I don't have an ETA on that. Please continue to depend on your fork of `bazel-distribution` for the time being, apologies for the inconvenience.

mgosk:
+1 for merging this PR into master branch. I am already using this change from PR branch.

PR #372, opened 2023-03-06
Extend NPM deployment with password or token-based deployment

flyingsilverfin:

As of January 2023, the `deploy_npm` using npm v8.x rule appears to stop working with pre-generated auth tokens for Sonatype repositories. However, we discover that the simpler mechanism of using base64 encoded username/password does work.

Using the base64 encoded authenciation is also [recommended](https://help.sonatype.com/repomanager3/nexus-repository-administration/formats/npm-registry/npm-security#npmSecurity-AuthenticationUsingBasicAuth) by Sonatype, we can successfully deploy.

However, the deployment to centralised NPM repositories may still require using token-based authentication. As a result, we extend the deployment rule to support both username/password or token-based authentication.

This issue was last worked on in https://github.com/vaticle/bazel-distribution/pull/330

* Introduce the capability to use username+password authentication OR token authentication based on which environment variables are present

### DIFF
--- a/npm/deploy/Deployer.kt
+++ b/npm/deploy/Deployer.kt
@@ -21,14 +21,16 @@
 
 package com.vaticle.bazel.distribution.npm.deploy
 
-import com.vaticle.bazel.distribution.common.Logging.Logger
 import com.vaticle.bazel.distribution.common.Logging.LogLevel.DEBUG
-import com.vaticle.bazel.distribution.common.OS.MAC
+import com.vaticle.bazel.distribution.common.Logging.Logger
 import com.vaticle.bazel.distribution.common.OS.LINUX
+import com.vaticle.bazel.distribution.common.OS.MAC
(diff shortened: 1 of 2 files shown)


npm/deploy/Deployer.kt
```diff
@@ -21,26 +21,49 @@
...
+                        arg("deploy_npm.tgz")),
+                env = mapOf("PATH" to pathEnv()))
+    }
+
+    private fun authURI(options: Options): String {
+        val token = options.npmToken;
+        val user = options.npmUsername
+        val pass = options.npmPassword
+        val uriPrefix = "--" + authParamFormattedURI(options.registryURL)
+        if (token != null) {
```
flyingsilverfin, line 55: If the token is there, username/pass cannot be provided, and vice versa. This is to make sure the usage is as unambiguous as possible.

PR #373, opened 2023-03-08
Fix pip deployment and extend deploy rule doc strings with usage

flyingsilverfin:

The `deploy_pip` rule was modified by #368 to allow deploying wheels and targz packages to repositories provided by a `pypirc` file. However, this broke this repository's consistent design: every `deploy_*` rule provides all possible repositories to upload to (eg. using `snapshot` or `release`), then selects which one to deploy to at run time using arguments.

This PR revert some changes to the `deploy_pip` rule to make it conform to this consistent design. We also fix some bugs during the upload process. 

Finally, we extend the doc strings for each of the deploy rules to indicate how to select the repository to deploy to. 

### `deploy_pip` changes

The `deploy_pip` rule is changed to take multiple different destination repositories to upload to. These are provided, as before, using the existing `snapshot` or `release` attributes. The additional attribute of `pypirc_repository` (renamed from `pypi_profile`) is made available in the same way, rather than with an exclusive-or logic.

Now we declare `deploy_pip` targets like this:

```
deploy_pip(
  name = "some-deploy-pip",
  snapshot = "https://... snapshot ...",
  release = "https:// ... release ...",
  pypirc_repository = "some-repository-name"
)
```

The `pypirc_repository` attribute refers the name of a repository declared in a [pyripc file](https://packaging.python.org/en/latest/specifications/pypirc/). This configuration file is an external file normally exists in a standard location such as `$HOME/.pypirc`.

After this change, deployment looks like this:
```
bazel run //:some-deploy-pip -- [pypirc|snapshot|release]
```

All deployment rules in this repository should continue to support declaring multiple target repositories in this manner, for the time being.

* Add doc strings to show how to use deployment rules for either snapshot or release
* Refactor `deploy_pip` to be consistent with the existing style, in which all potential deployment repositories are declared by attribute and the relevant one is selected as an argument to `bazel run //... -- [arg]`
* Fix the upload of the `.tar.gz` and `.whl` packages to the target repository, which was failing because of non-existent paths
* Downgrade `setuptools` to version `65.0.0`, which still allows non-semantic versioned packages to distributed
 and is a requirement for Vaticle for now

### DIFF
--- a/brew/rules.bzl
+++ b/brew/rules.bzl
@@ -114,5 +114,8 @@ deploy_brew = rule(
         "deployment_script": "%{name}.py"
     },
     implementation = _deploy_brew_impl,
-    doc = "Deploy Homebrew (Caskroom) formula to Homebrew tap"
+    doc = """Deploy Homebrew (Caskroom) formula to Homebrew tap.
+
+    Select deployment to `snapshot` or `release` repository with `bazel run //some-deploy-brew -- [snapshot|release]
+    """
 )
--- a/brew/rules.bzl
+++ b/brew/rules.bzl
@@ -116,6 +116,6 @@ deploy_brew = rule(
     implementation = _deploy_brew_impl,
     doc = """Deploy Homebrew (Caskroom) formula to Homebrew tap.
 
-    Select deployment to `snapshot` or `release` repository with `bazel run //some-deploy-brew -- [snapshot|release]
+    Select deployment to `snapshot` or `release` repository with `bazel run //:some-deploy-brew -- [snapshot|release]
     """
 )
--- a/pip/requirements.txt
+++ b/pip/requirements.txt
@@ -26,7 +26,8 @@ requests-toolbelt==0.10.1
 rfc3986==2.0.0
 rich==13.2.0
 secretstorage==3.3.3
-setuptools==66.1.1
+### DO NOT UPGRADE - setuptools > 65.x.x removes flexible versioning ###
+setuptools==65.0.0
 six==1.16.0
 tqdm==4.64.1
 twine==4.0.2
--- a/pip/rules.bzl
+++ b/pip/rules.bzl
@@ -173,7 +173,7 @@ def _deploy_pip_impl(ctx):
             "{package_file}": ctx.attr.target[PyDeploymentInfo].package.short_path,
             "{wheel_file}": ctx.attr.target[PyDeploymentInfo].wheel.short_path,
             "{version_file}": ctx.attr.target[PyDeploymentInfo].version_file.short_path,
-            "{pyprc_repository}": ctx.attr.pyprc_repository,
+            "{pypirc_repository}": ctx.attr.pypirc_repository,
             "{snapshot}": ctx.attr.snapshot,
             "{release}": ctx.attr.release,
         }
@@ -310,9 +310,9 @@ deploy_pip = rule(
             default = "",
             doc = "Release repository URL to deploy pip artifact to"
         ),
-        "pyprc_repository": attr.string(
+        "pypirc_repository": attr.string(
             default = "",
-            doc = "Repository name in the .pyprc profile to deploy to"
+            doc = "Repository name in the .pypirc profile to deploy to"
         ),
         "_deploy_py_template": attr.label(
             allow_single_file = True,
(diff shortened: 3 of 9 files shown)


brew/rules.bzl
```diff
@@ -114,5 +114,8 @@ deploy_brew = rule(
         "deployment_script": "%{name}.py"
     },
     implementation = _deploy_brew_impl,
-    doc = "Deploy Homebrew (Caskroom) formula to Homebrew tap"
+    doc = """Deploy Homebrew (Caskroom) formula to Homebrew tap.
+
+    Select deployment to `snapshot` or `release` repository with `bazel run //some-deploy-brew -- [snapshot|release]
```
benjaminasmall, line 119: Colon.

pip/requirements.txt
```diff
@@ -26,7 +26,8 @@ requests-toolbelt==0.10.1
 rfc3986==2.0.0
 rich==13.2.0
 secretstorage==3.3.3
-setuptools==66.1.1
+### DO NOT UPGRADE - setuptools > 65.x.x removes flexible versioning ###
+setuptools==65.0.0
```
flyingsilverfin, line 30: We can't ugprade to 66.0 because it removes non-semantic versioning support - https://github.com/pypa/setuptools/issues/3772#issuecomment-1384342813

pip/rules.bzl
```diff
@@ -164,14 +164,6 @@ def _assemble_pip_impl(ctx):
...
 def _deploy_pip_impl(ctx):
     deployment_script = ctx.actions.declare_file("{}_deploy.py".format(ctx.attr.name))
-    if ctx.attr.snapshot and ctx.attr.release:
-        fail(msg="cannot deploy snapshot and release at the same time.")
-    if ctx.attr.snapshot and ctx.attr.pypi_profile:
-        fail(msg="cannot deploy using pypi_profile and snapshot url at the same time.")
-    if ctx.attr.release and ctx.attr.pypi_profile:
-        fail(msg="cannot deploy using pypi_profile and release url at the same time.")
-    if not ctx.attr.pypi_profile and not (ctx.attr.snapshot or ctx.attr.release):
-        fail(msg="either pypi_profile or one of the snapshot or realease url is needed.")
```
flyingsilverfin, line 174: we do want to be able to provide multiple repositories at a time

pip/rules.bzl
```diff
@@ -181,7 +173,7 @@ def _deploy_pip_impl(ctx):
             "{package_file}": ctx.attr.target[PyDeploymentInfo].package.short_path,
             "{wheel_file}": ctx.attr.target[PyDeploymentInfo].wheel.short_path,
             "{version_file}": ctx.attr.target[PyDeploymentInfo].version_file.short_path,
-            "{pypi_profile}": ctx.attr.pypi_profile,
```
flyingsilverfin, line 184: `pypi` is not a term in the python packaging standards, this actually intends to refer to a `pyprc` repository name

pip/templates/deploy.py
```diff
@@ -30,48 +30,67 @@
...
+repositories = {
+    PYPRC_KEY: "{pyprc_repository}",
+    SNAPSHOT_KEY: "{snapshot}",
+    RELEASE_KEY: "{release}"
+}
+
+parser = argparse.ArgumentParser()
+parser.add_argument('repo_type')
+
+def upload_command(repo_type_key, package_file, wheel_file):
```
flyingsilverfin, line 50: Packaged the command generation into a method that takes args about where to upload to and which files to upload

pip/templates/deploy.py
```diff
@@ -30,48 +30,67 @@
...
+        return [package_file, wheel_file, '-u', pip_username, '-p', pip_password, '--repository-url', repositories[repo_type_key]]
+    else:
+        raise Exception(f"Unrecognised repository selector: {repo_type_key}")
+
+
+if not os.path.exists("{package_file}"):
+    raise Exception("Cannot find expected distribution .tar.gz to deploy at '{package_file}'")
+
+if not os.path.exists("{wheel_file}"):
+    raise Exception("Cannot find expected distribution wheel to deploy at '{wheel_file}'")
```
flyingsilverfin, line 71: pre-validate the `.tar.gz` and `wheel` pacakges exist before proceeding

pip/templates/deploy.py
```diff
@@ -30,48 +30,67 @@
...
+    new_package_file = dist_dir + "/{package_file}".replace(".tar.gz", "-{}.tar.gz".format(version))
+    new_wheel_file = dist_dir + "/{wheel_file}".replace(".whl", "-{}.whl".format(version))
+
+    if not os.path.exists(os.path.dirname(new_package_file)):
+        os.makedirs(os.path.dirname(new_package_file))
 
-    if os.path.exists("{package_file}"):
-        shutil.copy("{package_file}", new_package_file)
+    if not os.path.exists(os.path.dirname(new_wheel_file)):
+        os.makedirs(os.path.dirname(new_wheel_file))
```
flyingsilverfin, line 89: Make paths that may not exist - the `{package_file}` contains its own sub-directories such as `grpc/python/python-protocol.tar.gz` that are not going to exist under `dist/` unless we create them

pip/rules.bzl
alexjpwalker: Rename to `pypirc`

alexjpwalker (approved):

PR #375, opened 2023-03-09
Use PEP491 naming conventions for wheel deployments to pip

flyingsilverfin:

The new functionality introduced in #368 to deploy wheels did not introduce the right naming convention for wheels - PyPi for example will decline packages not naming according to the [PIP491](https://peps.python.org/pep-0491/).

This PR restricts the `assemble_pip` to only produce `py3` type archives, which can then be deployed with the required naming convention.

* Restrict `assemble_pip` to only allow Python-3 compatibility
* `deploy_pip` produces wheels that conform to `{name}-{version}-{pyversion}-{abi}-{platform}` by hardcoding it to `{name}-{version}-py3-none-any`

### DIFF
--- a/pip/templates/deploy.py
+++ b/pip/templates/deploy.py
@@ -77,17 +77,17 @@ with open("{version_file}") as version_file:
     version = version_file.read().strip()
 try:
     new_package_file = dist_dir + "/{package_file}".replace(".tar.gz", "-{}.tar.gz".format(version))
-    new_wheel_file = dist_dir + "/{wheel_file}".replace(".whl", "-{}.whl".format(version))
+    new_wheel_pep491 = dist_dir + "/{wheel_file}".replace("-", "_").replace(".whl", "-{}-py3-none-any.whl".format(version))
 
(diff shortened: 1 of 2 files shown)


pip/templates/deploy.py
```diff
@@ -77,17 +77,17 @@ def upload_command(repo_type_key, package_file, wheel_file):
     version = version_file.read().strip()
 try:
     new_package_file = dist_dir + "/{package_file}".replace(".tar.gz", "-{}.tar.gz".format(version))
-    new_wheel_file = dist_dir + "/{wheel_file}".replace(".whl", "-{}.whl".format(version))
+    new_wheel_pep491 = dist_dir + "/{wheel_file}".replace("-", "_").replace(".whl", "-{}-py3-none-any.whl".format(version))
```
flyingsilverfin, line 80: note that they also require `_` to be used for wheel package name itself, instead of `-`. The suffixes must be `-` separated.

alexjpwalker (approved):

PR #381, opened 2023-08-03
Emit canonical dependency info when assembling crate

dmitrii-ubskii:

We improve the generated crate manifest by allowing to specify available crate features, and by enabling the features of dependencies.

We did not include enabled features in the crate dependencies before this change. Now, we parse the crate universe manifests and collect the `crate_features` attributes from the `rust_library` dependencies of the crate being assembled so that we can emit the correct dependency requirement in the generated Cargo.toml. 

`rules_rust` must resolve the full list of features which were enabled explicitly or implicitly in order to correctly compile the dependency. That has the consequence of us ultimately explicitly depending on implementation details of other crates (see e.g. https://github.com/vaticle/typedb-driver-rust/issues/35), which is misleading on the crates.io page, and breaks across dependency versions. For this reason, we cannot simply rely on the `crate_features` attribute of the generated dependency files, but must instead parse the explicitly enabled features in the universe manifests.

### DIFF
--- a/crates/CrateAssembler.kt
+++ b/crates/CrateAssembler.kt
@@ -99,8 +99,8 @@ class CrateAssembler : Callable<Unit> {
     @Option(names = ["--repository"], required = true)
     lateinit var repository: String
 
-    @Option(names = ["--features"], split = ";")
-    var features: Array<String> = arrayOf()
+    @Option(names = ["--crate-features"], split = ";")
+    var crateFeatures: Array<String> = arrayOf()
 
     @Option(names = ["--universe-manifests"], split = ";")
     var universeManifests: Array<File> = arrayOf()
@@ -169,7 +169,7 @@ class CrateAssembler : Callable<Unit> {
             set<String>("path", crateRootPath)
         }
 
-        val canonicalDeps = universeManifests.flatMap {
+        val universeDeps = universeManifests.flatMap {
             TomlParser().parse(it.inputStream())
                     .getOrElse("dependencies", Config.inMemory()).entrySet().asSequence()
         }.associate { it.key to it.getValue<String>() }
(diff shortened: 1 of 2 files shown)


crates/CrateAssembler.kt
```diff
@@ -157,12 +168,40 @@ class CrateAssembler : Callable<Unit> {
...
-                set<String>(dep, ver)
+                if (canonicalDeps.contains(dep))
+                    set<String>(dep, canonicalDeps[dep])
+                else if (bazelDepFeatures.containsKey(dep)) {
+                    val depConfig = Config.inMemory()
+                    set<Config>(dep, depConfig)
+                    depConfig.set<String>("version", "=$ver")
+                    depConfig.set("features", bazelDepFeatures[dep])
+                } else
+                    set<String>(dep, "=$ver")
```
dmitrii-ubskii, line 191: Note: we _are_ going against the grain here by pinning the version in this way and not allowing it to vary. This can become a problem for the end user, especially if they are depending on another crate with this quirk: they would have to compile e.g. uuid v1.1.2, v1.1.3, and v1.1.4, because various libraries expect the exact versions of all, rather than being able to just depend on the highest SemVer-compatible version. This bloats the size of the binary, and greatly increases the compile and link times.

crates/CrateAssembler.kt
```diff
@@ -94,6 +99,12 @@ class CrateAssembler : Callable<Unit> {
     @Option(names = ["--repository"], required = true)
     lateinit var repository: String
 
+    @Option(names = ["--features"], split = ";")
```
flyingsilverfin, line 102: `--crate-features`

crates/CrateAssembler.kt
```diff
@@ -157,12 +168,40 @@ class CrateAssembler : Callable<Unit> {
             cargoToml.set<Config>("lib", this)
             set<String>("path", crateRootPath)
         }
+
+        val canonicalDeps = universeManifests.flatMap {
```
flyingsilverfin, line 172: `universeDeps`

crates/CrateAssembler.kt
flyingsilverfin: Note: discussion with the rest of the engineering team concludes that pinning the exact version all the way down is desirable for us, since we prioritise repeatable builds over user applications' compile times & a bit of size.

flyingsilverfin (approved):

PR #382, opened 2023-08-10
Update bazel rule sets and allow python packaging to include generated files

dmikhalin:

We upgrade rule sets `rules_python` and `rules_pkg`, and remove unnecessary workarounds related to documentation generation with `stardoc`.

We also fix python packaging to allow including generated library files. Generated library files for python are automatically put into the `bazel-out/.../bin` directory and the module can't reach them. We move these files into the module directory.

- We changed the paths of all files in `bazel-out/.../.../**/*` to `**/*`.
- We updated the `rules_python` version to 0.24.0 and replaced deprecated `pip_install()` with `pip_parse()`.
- Update `rules_pkg` to `0.9.1` and `rules_rust` to `v0.26.0`.


flyingsilverfin:
We have rolled in the changes from https://github.com/vaticle/bazel-distribution/pull/378 into this PR, except the nodejs permissioning error fix, which should be easy enough to re-create.

common/deps.bzl
```diff
@@ -5,28 +5,14 @@ def rules_python():
...
-            # generation.
-            "@vaticle_bazel_distribution//:bazelbuild_rules_python-export-requirements-bzl-for-stardoc.patch",
-        ],
-        patch_args = ["-p1"],
+        tag = "0.24.0",
     )
 
 def rules_pkg():
-    http_archive(
+    git_repository(
```
flyingsilverfin, line 12: Note: as described additionally in https://github.com/vaticle/bazel-distribution/pull/378
We need to depend on the git repository so that we get access to stardoc rules, in addition to the rest of the rules.

PR #387, opened 2023-09-28
Support architecture specific brew formulae

dmitrii-ubskii:

We add limited support for multiple homebrew artifact URLs with corresponding checksums.

- revert the checksum rule so that it computes the sha256 of the entire file rather than unzipping the file and checksumming its contents;
- move the platform configs from dependencies into here to break the circular dependency.

### DIFF
--- a/brew/rules.bzl
+++ b/brew/rules.bzl
@@ -33,18 +33,6 @@ def _deploy_brew_impl(ctx):
         else:
             substitution_files[key] = file.files.to_list()[0]
 
-    ctx.actions.expand_template(
-        template = ctx.file._deploy_brew_template,
-        output = ctx.outputs.deployment_script,
-        substitutions = {
-            '{brew_folder}': brew_formula_folder,
-            '{snapshot}' : ctx.attr.snapshot,
-            '{release}' : ctx.attr.release,
-            '{substitution_files}': json.encode({key: file.path for key, file in substitution_files.items()}),
-        },
-        is_executable = True,
-    )
-
     if not ctx.attr.version_file:
         version_file = ctx.actions.declare_file(ctx.attr.name + "__do_not_reference.version")
         version = ctx.var.get('version', '0.0.0')
@@ -56,6 +44,20 @@ def _deploy_brew_impl(ctx):
         )
     else:
         version_file = ctx.file.version_file
+
+    ctx.actions.expand_template(
+        template = ctx.file._deploy_brew_template,
+        output = ctx.outputs.deployment_script,
+        substitutions = {
+            '{brew_folder}': brew_formula_folder,
+            '{snapshot}' : ctx.attr.snapshot,
+            '{release}' : ctx.attr.release,
+            '{substitution_files}': json.encode({key: file.short_path for key, file in substitution_files.items()}),
+            '{formula_template}': ctx.file.formula.short_path,
+            '{version_file}': version_file.path,
+        },
+        is_executable = True,
+    )
 
     files = [
         ctx.file.formula,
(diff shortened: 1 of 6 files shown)


platform/BUILD
dmitrii-ubskii, line 1: Moved here from dependencies to break cyclic dependency.

platform/jvm/rules.bzl
```diff
@@ -367,9 +368,9 @@ def assemble_jvm_platform(name,
         main_class = main_class,
         jdk = jdk,
         os = select({
-            "@vaticle_dependencies//util/platform:is_mac": "Mac",
-            "@vaticle_dependencies//util/platform:is_linux": "Linux",
-            "@vaticle_dependencies//util/platform:is_windows": "Windows",
+            "@vaticle_bazel_distribution//platform:is_mac": "Mac",
+            "@vaticle_bazel_distribution//platform:is_linux": "Linux",
+            "@vaticle_bazel_distribution//platform:is_windows": "Windows",
```
dmitrii-ubskii, line 373: bazel-distribution should not depend on vaticle-dependencies

platform/jvm/rules.bzl
```diff
@@ -307,12 +307,13 @@ _assemble_zip_to_jvm_platform = rule(
...
+def native_jdk17():
     return select({
-        "@vaticle_dependencies//util/platform:is_mac": "@jdk16_mac//file",
-        "@vaticle_dependencies//util/platform:is_linux": "@jdk16_linux//file",
-        "@vaticle_dependencies//util/platform:is_windows": "@jdk16_windows//file",
+        "@vaticle_bazel_distribution//platform:is_linux_arm64": "@jdk17_linux_arm64//file",
+        "@vaticle_bazel_distribution//platform:is_linux_x86_64": "@jdk17_linux_x86_64//file",
+        "@vaticle_bazel_distribution//platform:is_mac_arm64": "@jdk17_mac_arm64//file",
+        "@vaticle_bazel_distribution//platform:is_mac_x86_64": "@jdk17_mac_x86_64//file",
+        "@vaticle_bazel_distribution//platform:is_windows_x86_64": "@jdk17_windows_x86_64//file",
```
dmitrii-ubskii, line 316: This is technically still a cycle — `@jdk17_*` are defined in dependencies, but they can come from anywhere. We are depending on a specific naming convention though.

common/checksum/rules.bzl
```diff
@@ -21,7 +21,7 @@ def _checksum(ctx):
     ctx.actions.run_shell(
         inputs = [ctx.file.archive],
         outputs = [ctx.outputs.checksum_file],
-        command = 'mkdir tmp; unzip {} -d tmp; shasum -a 256 tmp/* > {}; rm -rf tmp/'.format(ctx.file.archive.path, ctx.outputs.checksum_file.path)
+        command = 'shasum -a 256 {} | cut -d" " -f1 > {}'.format(ctx.file.archive.path, ctx.outputs.checksum_file.path)
```
dmitrii-ubskii, line 24: Reverted to getting the checksum of the entire archive rather than its contents.

brew/rules.bzl
```diff
@@ -46,34 +45,45 @@ def _deploy_brew_impl(ctx):
...
-            allow_single_file = True,
-            doc = 'Checksum of deployed artifact'
+        "file_substitutions": attr.label_keyed_string_dict(
+            allow_files = True,
+            doc = """
+            Substitute file contents into the formula.
+            Key: file to read the substitution from.
+            Value: placeholder in the formula template to substitute.
+            """
         ),
```
dmitrii-ubskii, line 87: We now allow for multiple checksums (different artifacts for different platforms), so this is now generic and allows any number of substitutions.

PR #405, opened 2024-03-21
Add version to uploaded apt package filename

krishnangovindraj:
The version is added to the uploaded filename of apt packages when being deployed. Without this, Cloudsmith's APT fetch ends up corrupted and downloads the wrong package, when multiple packages share the same file name across different versions.

* Adds a line in the deployment script to upload `package.deb` as `package_version.deb` on cloudsmith


flyingsilverfin (approved):
Looks good!

PR #409, opened 2024-04-05
Upgrade Packer and increase GCP packer image disk size

jamesreprise:

We've upgraded Packer to 1.8.3 and increased the GCP packer image disk size to 50GB.

As above.

### DIFF
--- a/packer/deps.bzl
+++ b/packer/deps.bzl
@@ -22,14 +22,12 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 def deps():
     http_archive(
         name = "packer_osx",
-        url = "https://releases.hashicorp.com/packer/1.7.4/packer_1.7.4_darwin_amd64.zip",
-        sha256 = "f3faf9dce0cebdfc7abfcf70511c6230e0c0a5c499ca3478def81549ded91b20",
-        build_file_content = 'exports_files(["packer"])'
+        build_file_content = 'exports_files(["packer"])',
+        url = "https://releases.hashicorp.com/packer/1.8.3/packer_1.8.3_darwin_amd64.zip",
     )
 
     http_archive(
         name = "packer_linux",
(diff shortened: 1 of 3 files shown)


packer/deps.bzl
```diff
@@ -22,14 +22,12 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 def deps():
     http_archive(
         name = "packer_osx",
-        url = "https://releases.hashicorp.com/packer/1.7.4/packer_1.7.4_darwin_amd64.zip",
-        sha256 = "f3faf9dce0cebdfc7abfcf70511c6230e0c0a5c499ca3478def81549ded91b20",
-        build_file_content = 'exports_files(["packer"])'
+        build_file_content = 'exports_files(["packer"])',
```
lolski, line 25: Is it possible to keep the `sha256`field and update the value, rather than removing them completely?

packer/deps.bzl
```diff
@@ -22,14 +22,12 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
...
+        url = "https://releases.hashicorp.com/packer/1.8.3/packer_1.8.3_darwin_amd64.zip",
     )
 
     http_archive(
         name = "packer_linux",
-        url = "https://releases.hashicorp.com/packer/1.7.4/packer_1.7.4_linux_amd64.zip",
-        sha256 = "3660064a56a174a6da5c37ee6b36107098c6b37e35cc84feb2f7f7519081b1b0",
-        build_file_content = 'exports_files(["packer"])'
+        build_file_content = 'exports_files(["packer"])',
```
lolski, line 31: And here too

packer/deps.bzl
jamesreprise: Done!

packer/deps.bzl
jamesreprise: Also done.

PR #411, opened 2024-04-09
pass twine publish args in deploy_pip

mgosk:

Be able to pass twine args eg `--skip-existing` , "--non-interactive"

Pass twine publish args in `deploy_pip` function

Live tested on private repository together with other waiting PRs https://github.com/mgosk/bazel-distribution/tree/bzlmod-with-pypi

### DIFF
--- a/pip/rules.bzl
+++ b/pip/rules.bzl
@@ -184,7 +184,8 @@ def _deploy_pip_impl(ctx):
             "{snapshot}": ctx.attr.snapshot,
             "{release}": ctx.attr.release,
             "{distribution_tag}": ctx.attr.distribution_tag,
-            "{suffix}": ctx.attr.suffix,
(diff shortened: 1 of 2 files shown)


alexjpwalker:
hi @mgosk , thank you for the contribution!

This change would be more aligned with the other rules in `bazel_distribution` if the arguments supported by `twine` were each individually configurable in the rule's arguments.

Something like
```py
        "skip_existing": attr.bool(
            default = False,
            doc = "Continue uploading files if one already exists. (Only valid when uploading to PyPI. Other implementations may not support this.)"
        ),
```

alexjpwalker (commented):
Commented in main thread

krishnangovindraj:
`--skip-existing` could be really useful even for us.

PR #427, opened 2026-02-23
Migrate to Bazel 8

lolski:
## Motivation
Update Bazel version from 6.2 to 8.5.1. 

The upgrade is done in a backwards-compatible way, such that "upstream" repositories that are yet to be upgraded may depend on this repository. This is done by preserving a few Bazel 6 configs (WORKSPACE, etc) alongside the new Bazel 8 ones.  Once every repositories have been upgraded to Bazel 8, these files will be removed.

### Create MODULE.bazel

The new module declaration is organized into these sections:

1. Module ID and version
2. Common/util declarations needed by subsequent declarations
3. Build dependencies
4. Languages and toolchains: where we declare our Rust, Kotlin, Python, and other languages we use
5. Platform definitions for doing architecture and OS selections
6. Package managers: pip, maven
7. Assembly and deployments
8. Docs

We've also done syntactic changes to fix broken rules during the upgrade, here's a few examples:

- The deprecated `cfg = "host"` updated to `cfg = "exec"`
- `output_to_genfiles = True` removed in various places.
- `struct.to_json()` to `json.encode()`

The following files are kept for backwards compatibility, and will be removed when all repos have been 

- WORKSPACE: Deprecated with notice; updated dep versions to match MODULE.bazel; added rules_cc() call


.factory/automation.yml
```diff
@@ -4,6 +4,6 @@ build:
       image: typedb-ubuntu-22.04
       type: foreground
       command: |
-        bazel build //...
+        bazel build //... --extra_toolchains=@llvm_toolchain//:all -- -//:docs -//:docs.extract
```
lolski, line 7: Excluded docs since I can't yet get it to work

.bazelrc
```diff
@@ -0,0 +1,2 @@
+# Bzlmod is now the default (Bazel 7+)
```
lolski: TODO: check if it is needed

.bazelversion
```diff
@@ -1 +1 @@
-6.2.0
+8.0.0
```
dmitrii-ubskii: Can we jump straight to the latest Bazel 8 (8.5.1)?

WORKSPACE
```diff
@@ -17,16 +17,23 @@
...
 
-# Load @rules_python, @io_bazel_rules_kotlin and @rules_jvm_external
-load("//common:deps.bzl", "rules_python", "rules_kotlin", "rules_jvm_external", "rules_rust")
+# Load @rules_python, @io_bazel_rules_kotlin, @rules_jvm_external, @rules_cc and @rules_rust
+load("//common:deps.bzl", "rules_python", "rules_kotlin", "rules_jvm_external", "rules_cc", "rules_rust")
 rules_python()
 rules_kotlin()
 rules_jvm_external()
+rules_cc()
 rules_rust()
```
dmitrii-ubskii: Is `rules_cc` used anywhere? Why the new dependency?

MODULE.bazel
```diff
@@ -0,0 +1,171 @@
...
+# ===========================================================
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+bazel_dep(name = "bazel_skylib", version = "1.7.1")
+
+# ===========================================================
+# Build dependencies
+# ===========================================================
+
+bazel_dep(name = "toolchains_llvm", dev_dependency = True, version = "1.3.0")
```
dmitrii-ubskii, line 23: 1) Am I correct in assuming this is required for Rust? If so, shouldn't `bazel_dep("rules_rust")` bring that in?
2) Does this mean Bazel's C++ toolchain is finally fully hermetic?

dmitrii-ubskii: Re: (1), would we be able to configure the toolchain if it were only a transitive dependency?

common/deps.bzl
```diff
@@ -1,11 +1,16 @@
+# DEPRECATED: This file is for WORKSPACE mode only.
+# New projects should use MODULE.bazel (Bzlmod) instead.
+# These functions will be removed in a future release.
```
dmitrii-ubskii: Should this disclaimer also appear in e.g. `maven/deps.bzl`? Its contents seem to have been inlined into `MODULE.bazel` as well.

MODULE.bazel
```diff
@@ -0,0 +1,171 @@
...
+llvm = use_extension(
+    "@toolchains_llvm//toolchain/extensions:llvm.bzl",
+    "llvm",
+    dev_dependency = True,
+)
+llvm.toolchain(name = "llvm_toolchain", llvm_version = "18.1.8",)
```
dmitrii-ubskii: Nit
```suggestion
llvm.toolchain(name = "llvm_toolchain", llvm_version = "18.1.8")
```

MODULE.bazel
```diff
@@ -0,0 +1,171 @@
...
+bazel_dep(name = "rules_nodejs", version = "6.3.2")
+node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
+node.toolchain(name = "nodejs", node_version = "17.9.1",)
+use_repo(node, "nodejs")
+
+# ===========================================================
+# Platform definitions
+# ===========================================================
+
+bazel_dep(name = "platforms", version = "0.0.10")
```
dmitrii-ubskii, line 81: This is also new, isn't it? I'm not seeing where this is used, should it be replacing something in `//platform`?

dmitrii-ubskii (commented):
Mostly questions.

> The following files are kept for backwards compatibility [...]
> .bazelrc: Created with Bzlmod migration comment

No it wasn't 🤔
But even so, is `.bazelrc` even deprecated? I see no indication of that in the docs (https://bazel.build/versions/8.5.0/run/bazelrc)

.bazelversion
lolski: > Can we jump straight to the latest Bazel 8 (8.5.1)?

Yeah it should be 8.5.1. Fixed!

> No it wasn't 🤔

outdated description - turned out we didn't need it at all! hence the file has been deleted.

WORKSPACE
lolski: I had to add it as it's become a dependency to loading `rules_rust` - otherwise the build would fail.

common/deps.bzl
lolski: Good spot, fixed.

MODULE.bazel
lolski: > Am I correct in assuming this is required for Rust?

It's actually to make the Python/Pip build to work. 

> Does this mean Bazel's C++ toolchain is finally fully hermetic?

In theory yes, but in practice guess I'll find out once I fully resolve TypeDB Driver's C/C++ implementation.

MODULE.bazel
lolski: Fixed

MODULE.bazel
lolski: Yes it's now. In Bazel 8, targets such as `"@platforms//os:linux"` has to be explicitly declared in before we can use it. Otherwise, the declarations we wrote in `//platform/constraints.bzl` won't work.

maven/deps.bzl
```diff
@@ -16,8 +16,12 @@
 # under the License.
 #
 
+# DEPRECATED: This file is for WORKSPACE mode only.
+# New projects should use MODULE.bazel (Bzlmod) instead.
+# These functions will be removed in a future release.
+
 maven_artifacts = [
-    "com.eclipsesource.minimal-json:minimal-json",
+    "com.eclipsesource.minimal-json:minimal-jsonz",
```
dmitrii-ubskii: Typo?

dmitrii-ubskii (approved):

lolski:
Thank you for your contribution @agluszak !

PR #430, opened 2026-05-26
Speed up CI with various fixes

krishnangovindraj:
## Usage and product changes
We use a custom namespace for the maven dependencies repo of the distribution tools, preventing clashes with a protobuf dependency update. We also update the nuget publishing job to append the `0.0.0-` to the commit-sha, so bazel doesn't discard the cache because the `--define version` changed.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -109,6 +109,7 @@ use_repo(
 bazel_dep(name = "rules_jvm_external", version = "6.6")
 maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
 maven.install(
+    name = "bazel_distribution_maven",
     artifacts = [
         "com.eclipsesource.minimal-json:minimal-json:0.9.5",
         "com.electronwill.night-config:core:3.6.5",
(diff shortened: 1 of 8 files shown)


nuget/rules.bzl
```diff
@@ -66,7 +66,9 @@ def _parse_version(ctx):
     version = ctx.attr.version
     if not version:
         version = ctx.var.get("version", "0.0.0")
-
+    if len(version) == 40:
+        # this is a commit SHA, most likely
+        version = "0.0.0-{}".format(version)
```
krishnangovindraj, line 71: Saves us from restarting bazel with a different --define="version="

MODULE.bazel
```diff
@@ -124,7 +125,7 @@ maven.install(
     ],
     strict_visibility = True,
 )
-use_repo(maven, "maven")
+use_repo(maven, "bazel_distribution_maven")
```
krishnangovindraj, line 128: Prevents a clash with protobuf which declares `@maven`

flyingsilverfin (approved):

farost (approved):

